PR #184, opened 2022-09-21
New query params for metadata

asutula:
Moving away from `mode=list` to `unwrap=true&extract=true`.


asutula:
@sanderpick once merged, we'll need to update the url template in the contract, whatever the best way is to do that.

sanderpick (approved):
nice! should we be prioritizing making this change in the contract? or is the back compat support for `mode` working fine for now?

asutula:
The legacy support is working fine for now, no rush.

PR #233, opened 2022-10-11
Add The Garage functionality

dtbuchholz:


sanderpick:
i'll dig in here after meetings today! btw, `@tableland/evm@3.0.0` is out now and includes usable `SQLHelpers.sol`

dtbuchholz:
cool, will update that as well as the `block.timestamp` -> `block.number` swap. trying to resolve some small test coverage issue with an `instanbul` branch as part of that, too.

ethereum/contracts/TablelandRigs.sol
```diff
@@ -380,6 +641,7 @@ contract TablelandRigs is
 
     /**
      * @dev See {ERC721A-_beforeTokenTransfers}.
+     * Also, adds logic to block transfers while a Rig is being piloted.
```
dtbuchholz, line 644: It might be useful to also add an `_afterTokenTransfer` table insert that helps track Rig ownership. I.e., a simple `rigs_ownership_index` table with `rig_id` and `owner`'s address.

Cross posting some comments about it: [see here](https://github.com/tablelandnetwork/rigs/commit/1d97b32ba315758d3ddc2a8a22a9bf4ece9bbf1b#r86312956)

ethereum/contracts/ITablelandRigs.sol
```diff
@@ -31,6 +55,20 @@ interface ITablelandRigs {
...
+    // A Rig's pilot
+    // TODO alterted this to align with the table column definitions
+    struct RigPilot {
+        // Index of the current pilot, for tracking the history of a Rig's pilots.
+        uint16 index;
+        // Keep track of the pilot's starting `block.number` for flight time tracking.
+        uint64 startTime;
+        // Address of the ERC721 pilot contract.
+        address pilotContract;
+        // TODO we have 160 + 64 + 16 = 240 bits in this tighly packed group, so room for 16 bits of extra data, if needed
```
dtbuchholz, line 67: In case we can find something useful with 16 bits -- calling this out.

ethereum/contracts/TablelandRigs.sol
```diff
@@ -49,13 +55,41 @@ contract TablelandRigs is
...
+
+    // Table ID for the Rigs pilot sessions table.
+    // TODO do we want to have a method for updating this value?
+    uint256 private _rigPilotSessionsTableId;
+
+    // Tracks the Rig `tokenId` to its current `RigPilot`.
+    mapping(uint256 => RigPilot) internal _pilots;
+
+    // Track piloted vs. parked Rigs as a bitlist.
+    // TODO this was implemented for gas reduction purposes while iterating/checking for duplicates in `_pilots`
```
dtbuchholz, line 82: Calling this out in case there are better ways to handle this optimization; `_rigsStatus` is a bitlist that tracks active Rigs. It's used for iteration purposes instead of accessing a `RigPilot.startTime` in the `_pilots` mapping, which also indicates if a Rig is in flight vs. parked.

It's still a gas guzzler down in `pilotRig` since it has to check every Rig for pilot duplicates, but doing this dropped around 5M gas (7.5M down to 2.5M)

ethereum/contracts/TablelandRigs.sol
```diff
@@ -354,6 +390,231 @@ contract TablelandRigs is
...
+        // been ungaraged yet since parking handles training logic (see `parkRig` for more details)
+        if (_pilots[tokenId].index == 0) revert RigIsNotTrained(tokenId);
+        // Check if Rig is currently in-flight (not parked)
+        if (_pilots[tokenId].startTime > 0) revert RigIsNotParked(tokenId);
+        // Check if `pilotContract` is an ERC721; cannot be the Rigs contract
+        // TODO is this the right way to handle checking if the ERC721 is the Rigs contract? Or, is there a different way w/upgradeable contracts
+        if (
+            !pilotContract.supportsInterface(type(IERC721).interfaceId) ||
+            (pilotContract.supportsInterface(type(IERC721).interfaceId) &&
+                pilotContract == address(this))
```
dtbuchholz, line 485: Note: was considering blocking `TablelandTables` from being a pilot as well.

ethereum/contracts/TablelandRigs.sol
dtbuchholz: Realizing this needs some significant overhaul...gonna think about it a bit more, but current prices mean $51 USD to park a Rig since we're on L1. Gonna check out that diamond storage impl vs. struct.

ethereum/contracts/TablelandRigs.sol
```diff
@@ -354,6 +390,231 @@ contract TablelandRigs is
...
+        for (uint256 i = 1; i <= maxSupply; i++) {
+            if (
+                // `_rigsStatus` starts at index `0`, while `tokenId` in `_pilots` starts at `1`
+                _rigsStatus[i - 1] == 1 &&
+                (_pilots[i].pilotContract == pilotContract &&
+                    _pilots[i].pilotId == pilotTokenId)
+            ) {
+                parkRig(i);
+            }
+        }
```
sanderpick, line 500: is this the bulk of the gas in this method? what about using two maps:

```
mapping(uint256 => uint16) internal _pilots; // pilot -> tokenid
mapping(uint16 => Pilot) internal _rigPilots; // tokenid -> pilot data
struct Pilot {
    uint16 index;
    uint64 startTime;
    uint256 pilot; // value stored twice
}
```

or

```
mapping(uint16 => uint256) internal _rigPilots; // tokenid -> pilot
mapping(uint256 => Pilot) internal _pilots; // pilot -> pilot data
struct Pilot {
    uint16 index;
    uint64 startTime;
    uint16 rigId;
}
```

both duplicate data, but second one duplicates less. 

used use `_rigPilots` to lookup a rig's pilot, and `_pilots` to look up whether or not a pilot has been used, and to get its data (start time, index)

ethereum/contracts/ITablelandRigs.sol
```diff
@@ -31,6 +55,20 @@ interface ITablelandRigs {
         PUBLIC
     }
 
+    // A Rig's pilot
+    // TODO alterted this to align with the table column definitions
+    struct RigPilot {
```
sanderpick, line 60: just `Pilot` would be okay too

ethereum/contracts/TablelandRigs.sol
sanderpick: dig it

ethereum/contracts/TablelandRigs.sol
```diff
@@ -354,6 +390,231 @@ contract TablelandRigs is
...
+        (bool isValidSubtraction, uint256 blockDifference) = SafeMathUpgradeable
+            .trySub(block.number, 172800);
+        if (
+            !isValidSubtraction ||
+            _pilots[tokenId].index == 0 ||
+            (_pilots[tokenId].index == 1 &&
+                !(blockDifference > _pilots[tokenId].startTime))
+        ) {
+            revert RigIsNotTrained(tokenId);
+        }
```
sanderpick, line 571: shouldn't we let rigs park even if training isn't finished? they could just start from 0 again

ethereum/contracts/TablelandRigs.sol
```diff
@@ -354,6 +390,231 @@ contract TablelandRigs is
...
+        _rigsStatus[tokenId - 1] = 0;
+        // Update the row in the `rig_pilot_sessions` table with its `end_time`
+        string memory setters = string.concat(
+            "start_time=0,end_time=(",
+            StringsUpgradeable.toString(uint64(block.number)),
+            "-",
+            StringsUpgradeable.toString(startTime),
+            ")"
+        );
+        // Only update the row with the matching `rig_id` and `start_time`
```
sanderpick, line 585: nice!

sanderpick (commented):
looking good! haven't looked at tests yet... just leaving some comments on contracts for now. will be playing with it more

ethereum/contracts/TablelandRigs.sol
sanderpick: i left a suggestion below about using two maps

ethereum/contracts/TablelandRigs.sol
dtbuchholz: @sanderpick yeah, that could work too...it basically ends up with the same result, minus the locking aspect, which makes it flexible for the user. we'd just need to make sure the frontend makes a callout to _not_ park the Rig if 30 consecutive days of training hasn't happened yet, otherwise, they'll start over (cc @datadanne).

dtbuchholz: one random idea for the extra data in the `Pilot` struct. we're currently tracking the `Pilot.index` of `1` to check if training has happened or not for a single Rig, and training only happens once. instead, we could maybe make training a requirement for every new Rig owner? namely, have simple `isTrained` field in the struct as a `uint8`, and on `_afterTokenTransfer`, this gets reset to `0`. idea being, the training isn't a one time thing, which makes it feel "nicer" as part of the code. just an idea.

(not sure if i originally misinterpreted the design, but this feels more natural too since a pilot should have to train to use a Rig)

dtbuchholz:
cool thanks -- and i have to update something with the tests anyways...forgot to clear the cache and broke the tests due to some missing artifacts.

sanderpick:
playing around with this, prob on your radar, but i had to add some contracts to the dependency compiler to get passing test in `hardhat.config.ts`:

```
dependencyCompiler: {
    paths: [
      "@tableland/evm/contracts/TablelandTables.sol",
      "@tableland/evm/contracts/test/TestSQLHelpers.sol",
      "@tableland/evm/contracts/test/TestAllowAllTablelandController.sol",
      "@tableland/evm/contracts/test/TestERC721Enumerable.sol",
      "@openzeppelin/contracts/finance/PaymentSplitter.sol",
    ],
  },
  ```

**Comment by dtbuchholz** [comment]
thanks, yeah, i had removed it for some reason, which obviously broke the tests. haven't impl that `pilotRig` optimization yet but cleaned up most of the other aspects.

File: ethereum/contracts/TablelandRigs.sol (hunk 5 above)
Comment (sanderpick) [reply ethereum/contracts/TablelandRigs.sol:502]: maybe clear already, but for an example of packing / unpacking multiple ints into a larger int, see `getClaimed` and `_setClaimed`:

Screen Shot 2022-10-15 at 11 18 34 AM

packs two `uint16`s into a `uint64`

Comment (sanderpick) [reply ethereum/contracts/TablelandRigs.sol:502]: Hrm, I think it needs to be the first flavor, since we can't index pilots that are the trainer in `mapping(uint256 => Pilot) internal _pilots;`

So, expanding on the first flavor, which can be (changing names to be more clear):

```
struct Pilot {
    uint64 startTime;
    uint192 pilot; // value is indexed by _pilotIindex
}

mapping(uint16 => Pilot) internal _pilots; // tokenid -> pilot data
mapping(uint192 => uint16) internal _pilotIndex; // packed pilot -> tokenid
```

...where "pilot" is pilot contract address (`uint160`) + pilot token id (`uint32`). `uint32` allows token ids up to `4,294,967,295`, which sounds like plenty to me. You could do a check on token id to see if it'll fit in `uint32`, and error if not.

Training should be as cheap as possible to get people started in the game. With this setup, training need not touch `_pilotIndex`, just `_pilots`. During training, we could say that `pilot` is `1`.

I think we can remove `index` entirely by saying that `_pilots` stores the "latest pilot data". ie, don't delete from `_pilots` on park, just flip `startTime` to `0`. Then you can walk through some logic to determine state: 

Possible states:
1. not trained
2. in training
3. trained and parked
4. trained and piloted

- It's possible to go from (2) to (1) by "parking" while training, ie, rig must start from zero again.
- It's possible to go from (4) to (3)—that's what "parking" does when already trained.

You might consider creating a method that returns "garage status" as an enum to make logic in methods a bit cleaner. Given the possible states, the logic of this method would look something like:

```
enum GarageStatus {
    UNTRAINED,
    TRAINING,
    PARKED,
    PILOTED
}

function _pilotStatus() private returns (GarageStatus) {
    Pilot p = _pilots[uint16(tokenId)];
    if (p.startTime == 0) {
        if (p.pilot == 0) return UNTRAINED;
        // The `park` logic should set `pilot` to `0` if both of these are true:
        //   - `pilot` is currently `1`
        //   - rigs has not been training for long enough
        // ie, `park` results in a status of UNTRAINED or PARKED
        // In no other cases should `park` update `pilot` (it's either current or last used pilot)
        else return PARKED;
    } else {
        // Invariant: pilot cannot be `0` if `startTime` is > `0`.
        // ie, if rig is training or being piloted, pilot is either the trainer or a real pilot
        if (p.pilot == 1) return TRAINING;
        else return PILOTED;
    }
}

function _canPilot() private returns (bool) {
    // two ways to pilot:
    //   1. in PARKED status
    //   2. in TRAINING status for long enough
    // Invariant: you cannot switch real pilots mid-flight, but you can go
    // from trainer -> real pilot to make it easier + cheaper to pilot (continue playing the game)
    Pilot p = _pilots[uint16(tokenId)];
    if (p.startTime == 0) {
        // Rig is either UNTRAINED or PARKED. It's not possible to pilot from UNTRAINED.
        return p.pilot != 0; // could be trainer (`1`) or real pilot
    } else if (p.pilot == 1) {
        return block.number >= p.startTime + 172800;
    } else return false;
}
```

Comment (sanderpick) [reply ethereum/contracts/TablelandRigs.sol:502]: The only function of `_pilotIndex` here is to easily check if a custom pilot is in use.

File: ethereum/contracts/TablelandRigs.sol (hunk 7 above)
Comment (sanderpick) [reply ethereum/contracts/TablelandRigs.sol]: see https://github.com/tablelandnetwork/rigs/pull/233#discussion_r996351264. it think we can handle this in the way described there. i like the idea of "training the rig" vs. pilot (they are sentient machines :)), which adds some friction to getting flight time for people that have tons of rigs, vs. them being able to train on one rig for piloting many.

File: ethereum/contracts/TablelandRigs.sol [hunk 9]
```diff
@@ -354,6 +390,417 @@ contract TablelandRigs is
...
+        );
+        emit Training(tokenId);
+    }
+
+    using ERC165Checker for address;
+
+    /**
+     * @dev See {ITablelandRigs-pilotRig}.
+     */
+    function pilotRig(
```
dtbuchholz, line 610: @datadanne was thinking about the use case for pilot -> park -> pilot, where the same / most recent pilot is being used in the second piloting step.

with the second attempt, which would you expect to pass from the frontend:
1. `tokenId`, `pilotContract`, and `pilotId` -- i.e., pass the same values as the first piloting step
2. `tokenId`, `pilotContract` as `0x0`, and `pilotId` as `0`, and let the SC perform the necessary checks to pilot with the most recently used pilot
3. only `tokenId`, and let the SC perform the necessary checks

datadanne: Number 1! For 2 and 3 the client would have to keep a bunch of state and somehow recalculate that if the user reloads the page in between parking and piloting again

dtbuchholz:
Closing in favor of rebased branch at https://github.com/tablelandnetwork/rigs/pull/284

PR #245, opened 2022-10-17
Support trainer pilot, use tableland for metadata

sanderpick:


animation_url/.prettierrc
```diff
@@ -1,9 +1,7 @@
 {
-	"useTabs": true,
```
sanderpick, line 2: @joewagner this was the culprit

animation_url/src/routes/[slug]/+page.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  return {
+    imageUrl: httpUri,
+    badges: [],
+    pilot,
+  };
+}
+
+/* NOTE: badge handling logic to be used later
```
Comment (sanderpick) [inline animation_url/src/routes/[slug]/+page.ts:45]: added this badge handling logic back in, but commented out

sanderpick:
vercel bot is pretty slick. i can move the deploy to the main branch when this makes it there

animation_url/.prettierrc
joewagner: Ah! that makes sense

animation_url/src/routes/[slug]/+page.ts
```diff
@@ -0,0 +1,80 @@
...
+const ipfsGatewayUri = "https://nftstorage.link/ipfs/";
+
+/** @type {import('./$types').PageLoad} */
+export async function load({ url }) {
+  const tableland = connect({ chain: "polygon-mumbai" });
+
+  // get rig id, allowing for .html suffix
+  let rigId = url.pathname;
+  const parts = rigId.replace(/^\/|\/$/g, "").split(".");
+  rigId = parts[0];
```
Comment (joewagner) [inline animation_url/src/routes/[slug]/+page.ts:15]: Would it make sense to check if we got the rigId here and throw if it's not what we expect?

joewagner (approved):
I left a small question/nit inline, but looks great!

animation_url/src/routes/[slug]/+page.ts
Comment (sanderpick) [reply animation_url/src/routes/[slug]/+page.ts:15]: could yeah, `rigId` _will_ be something at this point, but could be a garbage route. i was just letting the sdk return an error from the validator since any true validation would be involved, ie, check it's a number, check it's within range 1-3000. 

Screen Shot 2022-10-17 at 3 44 16 PM

Comment (joewagner) [reply animation_url/src/routes/[slug]/+page.ts:15]: That makes sense

PR #270, opened 2022-10-31
feat: pack pilots

dtbuchholz:
Implementation of a single `uint256` packed pilot, instead of a previously implemented Pilot struct. The former contains packing of all pilot info (`startTime`, `pilotId`, and `pilotContract`) in a single `uint256`, and the latter had contained `startTime` and `pilot`, where the `pilot` was packing `pilotId`, and `pilotContract`.

### DIFF
--- a/ethereum/contracts/TablelandRigs.sol
+++ b/ethereum/contracts/TablelandRigs.sol
@@ -557,12 +557,13 @@ contract TablelandRigs is
         // There are two ways to pilot:
         //   1. In a `PARKED` status
         //   2. In a `TRAINING` status for long enough (30 days; 172800 blocks)
-        if (_startTime(uint16(tokenId)) == 0) {
+        uint64 startTime = _startTime(uint16(tokenId));
+        if (startTime == 0) {
             // Rig is either `UNTRAINED` or `PARKED`
             // It's not possible to pilot from `UNTRAINED`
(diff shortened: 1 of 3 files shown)


ethereum/contracts/TablelandRigs.sol
```diff
@@ -54,16 +70,19 @@ contract TablelandRigs is
...
 
-    // Tracks the Rig `tokenId` to its current `Pilot`.
-    mapping(uint16 => Pilot) internal _pilots;
+    // Tracks the Rig `tokenId` to its current pilot, represented as a packed `uint256`.
+    //
+    // Bits layout:
+    // - [0..63]    `startTime` - starting block number of pilot's flight time
+    // - [64..95]   `pilotId` - ERC-721 token ID of the pilot at `pilotContract`
+    // - [96..255]  `pilotContract` - address of the ERC-721 contract for the pilot
+    mapping(uint16 => uint256) internal _pilots;
```
dtbuchholz, line 82: @sanderpick i was hacking around with packing pilot data into a `uint256`, instead of representing pilots as a `Pilot` struct (which contained the start time and a packed `uint192` variable of pilot ID and contract). i based it off of the ERC721A impl.

it didn't really have a significant gas savings, so i'm curious if you think it's worthwhile to use this impl, or this all hurts readability.

PR #284, opened 2022-11-02
Add pilots

dtbuchholz:
Adds piloting / The Garage functionality to the Rigs smart contract and associated tests and deployment changes.


asutula:
I see that `artifacts/local.db` has some changes. There should't be any changes there, or let me know why there is. If I'm correct, you should revert that change.

asutula (changes requested):
I see that artifacts/local.db has some changes. There should't be any changes there, or let me know why there is. If I'm correct, you should revert that change.

dtbuchholz:
Closing in favor of https://github.com/tablelandnetwork/rigs/pull/285, which maintains verified commits and fixes the `artifacts/local.db` issue noted above.

PR #288, opened 2022-11-03
Fix SIWE implementation

datadanne:
Replaces the hacky use of the tableland sdk siwe call in the connect button with a proper `RainbowKitAuthenticationProvider`.

Related, we should maybe look at refactoring the SDK code so that we can re-use the code for auth implementations like this? I had to copy/paste the code we use in the sdk to create the message since we don't expose that code.

### DIFF
--- a/garage/package-lock.json
+++ b/garage/package-lock.json
@@ -20,6 +20,7 @@
         "react": "^18.2.0",
         "react-dom": "^18.2.0",
         "react-router-dom": "^6.4.1",
+        "siwe": "^2.1.3",
         "wagmi": "^0.6.7"
       },
       "devDependencies": {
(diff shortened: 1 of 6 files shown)


joewagner:
> Related, we should maybe look at refactoring the SDK code so that we can re-use the code for auth implementations like this? I had to copy/paste the code we use in the sdk to create the message since we don't expose that code.

This brings up something I noticed as an outcome of the season 5 planning. When we [remove the validator relay,](https://github.com/tablelandnetwork/js-tableland/issues/235)  write queries will go to the chain, and when we [switch to RESTful api endpoints](https://www.notion.so/textile/We-should-drop-support-for-JSON-RPC-and-focus-only-on-REST-APIs-7dcad940106e4340bd4892b4c10a31c2) the chain id will be included in the path when needed.  Which means the SDKs will no longer need to generate a siwe token.

sanderpick (approved):
nice, lgtm!

sanderpick (approved):

PR #314, opened 2022-11-07
Add missing pilotSessionsTable to deployment definitions

datadanne:
This was incorrectly removed in a1b3d6d77827ff93bcaa1da707ccd82bca871e5f


sanderpick (approved):
nice, i think the pilots table was removed because it's queryable from the contract, but may as well hardcode too.

PR #317, opened 2022-11-07
Fix builds, use latest contract deployment & remove badges section with dummy data

datadanne:
- Fixes a dependency issue introduced by dependabot in https://github.com/tablelandnetwork/rigs/pull/311 where wagmi was updated to v0.7.x which is incompatible with our version of rainbowkit. There is a breaking change in the wagmi upgrade v0.6 -> v0.7, I'm gonna work on upgrading the app to use v0.7
- Reads deployment info directly from `ethereum/deployments.ts` instead of the npm package
- Removes the badges section with dummy data from the rig details screen

### DIFF
--- a/garage/package-lock.json
+++ b/garage/package-lock.json
@@ -21,7 +21,7 @@
         "react-dom": "^18.2.0",
         "react-router-dom": "^6.4.3",
         "siwe": "^2.1.3",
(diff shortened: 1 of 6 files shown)


sanderpick (approved):
lgtm!

PR #342, opened 2022-11-08
Publish layers and renders fix

asutula:
Ended up being a stupid bug in processing and using cli flags then properly detecting the empty value of a `cid.Cid`. Also added some flags to make nft.storage upload concurrency and rate limit configurable.

### DIFF
--- a/cmd/rigs/cmd/publish_layers.go
+++ b/cmd/rigs/cmd/publish_layers.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/ipfs/go-cid"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


cmd/rigs/cmd/publish_layers.go
```diff
@@ -32,17 +35,22 @@ var layersCmd = &cobra.Command{
 		}
 		chunksDir := viper.GetString("chunks-dir")
 
-		if chunksDir == "" && c.String() == "" {
+		if chunksDir == "" && !c.Defined() {
```
asutula, line 38: This was the bug. The empty cid string is `b`. Better to use the `Defined()` func. So this was causing the step of adding the data to IPFS to be skipped.

sanderpick (approved):
sweet

PR #348, opened 2022-11-09
Garage: Include Andale Mono font files to support devices without Andale installed

datadanne:
@sanderpick question: Do we need to buy a license to use the font or do we already have one from the main website that allows us to use the font here as well?


sanderpick:
actually we don't have a license. the font was chosen by a past employee... i wasn't aware that it isn't free. i'll grab a web license.

sanderpick (approved):

sanderpick:
(i grabbed one... i'll add the license after this is merged)

PR #351, opened 2022-11-09
Track gas price in addition to gas used

asutula:
To clarify, the `gas` column of the `txns` table is the units of gas used, and the `gas_price` column is the cost of each unit of gas in wei (not gwei; 1 gwei = 1,000,000,000 wei).


asutula:
Sample query to get each transaction's total fee in ETH:
```
select hash, gas, gas_price, gas*gas_price*1e-18 as txn_fee_eth from txns;
```

sanderpick (approved):
sweet

PR #364, opened 2022-11-10
Garage: Read zora api key from env var

datadanne:


garage/src/hooks/useZDK.ts
```diff
@@ -9,7 +9,7 @@ const API_ENDPOINT = "https://api.zora.co/graphql";
 const args = {
   endPoint: API_ENDPOINT,
   networks: [networkInfo],
-  // apiKey: process.env.API_KEY,
+  apiKey: import.meta.env.VITE_ZORA_API_KEY
```
sanderpick, line 12: 👍

sanderpick: can i use this for both dev and prod (mumbai and mainnet)?

sanderpick (approved):

garage/src/hooks/useZDK.ts
datadanne: Yeah we can for now. Zora's NFT api doesn't index mumbai but we don't need that feature right now, we will have to rethink the dev setup later if we want to load custom pilot NFTs

sanderpick: cool, makes sense

sanderpick:
gonna merge this one 🙏

PR #369, opened 2022-11-13
Updates local.db with publish data from arbitrum

sanderpick:
@asutula does is make sense that tableland.db is not touched?


asutula:
Yea, that makes sense. No changes to `artifacts/tableland.db` expected.

asutula (approved):

PR #383, opened 2022-11-15
Garage: add favicon & icons for touch devices

datadanne:
@sanderpick I added an all black version of the Tableland logo so that it would be different from the public website, do you thinks works or do have any other ideas?


joewagner (approved):
🚀

sanderpick:
nice! looks great

sanderpick (approved):

PR #405, opened 2022-11-16
Deployments tweak

sanderpick:
- replaces remaining goerli tables with arbitrum tables on mainnet
- fills in missing mumbai tables
- makes it easier to create needed tables on the correct network when deploying rigs to testnet / mainnet


joewagner (approved):
Says branch is out of date, but otherwise looks good!

PR #407, opened 2022-11-16
Use mainnet gateway

sanderpick:


sanderpick:
Verified that animation_url and garage deployments work fine.

PR #423, opened 2022-11-21
Explicit TEXT column for attribute value

### LINKED ISSUE #419: Use text column for rig_attributes values

asutula:
This fixes the Rigs data to align with the non-determinism fixes to be applied to the spec and parser. I pushed the new `rigs_attrbute` data to the `artifacts/tableland.db` as well as mumbai and arbitrum goerli, hence the updates to `artifacts/local.db` for tracking the new attributes table names and insert transactions.

We still need to create the new attributes table and write data to it on arbitrum mainnet. I don't have the access to do that, but will coordinate with @sanderpick and @andrewxhill  go get it done asap.

Closes #419


pkg/storage/tableland/common/sqlfactory.go
```diff
@@ -71,7 +70,12 @@ func (s *SQLFactory) SQLForInsertingRigAttributes(rigAttrTable string, rigs []lo
 		attVales = append(
 			attVales,
 			goqu.Vals{rig.ID, "string", "VIN", rig.VIN},
-			goqu.Vals{rig.ID, "number", "% Original", math.Round((rig.PercentOriginal90*100)*100) / 100},
+			goqu.Vals{
+				rig.ID,
+				"number",
+				"% Original",
+				strings.TrimRight(strings.TrimRight(fmt.Sprintf("%.2f", rig.PercentOriginal90*100), "0"), "."),
```
asutula, line 77: Here I'm formatting the `float64` value as a string, using up to two decimal places, but removing any trailing `0`s.

pkg/storage/tableland/store.go
```diff
@@ -44,7 +44,7 @@ var (
 			rig_id integer not null,
 			display_type text,
 			trait_type text not null,
-			value any not null,
+			value text not null,
```
asutula, line 47: No more `any`, using `text` explicitly.

brunocalza (approved):

PR #450, opened 2022-12-05
fix: pilot contract bugs & logic

dtbuchholz:
The `TablelandRigPilots.sol` contract had a SQL issues that included a missing comma and incorrect `end_time` logic. This fixes these as well as makes the associated changes to the tests, plus, a small fix on the verification script (adds error handling with already verified contracts, which will otherwise halt the verification process).

There were also some slight logic changes when changing a pilot, which now creates a new session instead of morphing an existing trainer session into a "real pilot" session. Lastly, the mumbai contract pilot contract was upgraded to test these logic changes.

### DIFF
--- a/ethereum/.openzeppelin/polygon-mumbai.json
+++ b/ethereum/.openzeppelin/polygon-mumbai.json
@@ -4529,6 +4529,284 @@
           }
         }
       }
+    },
+    "a69c1538e7eb6ddf0b6e6cb5decad09e84bdd46fec60b6c747f6cc4f80d6220a": {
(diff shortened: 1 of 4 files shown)


datadanne (approved):
lgtm! should we make the change so that force parked-rigs get 0 FT for the ongoing session now as well since we are going to start force parking soon?

dtbuchholz:
I can definitely add that! I wanted to keep it to the basics since the existing logic _will_ reset the FT upon parking while still in a `TRAINING` state. But it's not a huge lift to include that.

PR #461, opened 2022-12-06
feat: remove session FTR on force park

dtbuchholz:
- Changed `parkRig` logic to check if the caller is force parking the Rig. Namely, if the `sender` passed into `parkRig` is not the token owner, that means the owner of the Rigs parent contract / admin called `parkRig`. If this is true, remove FTR earned (i.e., set `start_time=end_time`). Previously, a force park via `parkRigAsOwner` resulted in FTR accrual, unless the RIg happened to still be in a "training" status.
- Updated associated `parkRig` tests.


ethereum/contracts/TablelandRigPilots.sol
```diff
@@ -386,29 +387,35 @@ contract TablelandRigPilots is
     /**
      * @dev See {ITablelandRigPilots-parkRig}.
      */
-    function parkRig(address /*sender*/, uint256 tokenId) public onlyParent {
+    function parkRig(address sender, uint256 tokenId) public onlyParent {
```
dtbuchholz, line 390: fwiw, 1.5k gas increase

dtbuchholz:
Closing in favor of https://github.com/tablelandnetwork/rigs/pull/464

PR #462, opened 2022-12-06
Enable viewing pilots in animation_url app

joewagner:


animation_url/src/routes/[slug]/+page.ts
```diff
@@ -7,14 +8,31 @@ const isDev =
...
 const ipfsGatewayUri = isDev
   ? "https://nftstorage.link/ipfs/"
   : "https://tableland.mypinata.cloud/ipfs/";
 
+// NOTE: zdk docs say that only mainnet is supported currently. We could be
+//       supporting testnet here, but for the time being testnet localnet
+//       collections will show the pilot from mainnet.
+const networkInfo = {
+  network: ZDKNetwork.Ethereum,
+  chain: ZDKChain.Mainnet,
```
Comment (joewagner) [inline animation_url/src/routes/[slug]/+page.ts:21]: The [zora labs zdk docs](https://www.npmjs.com/package/@zoralabs/zdk) mention that they only support mainnet ethereum.  This means only mainnet ethereum NFT pilots can be viewed.

animation_url/src/routes/[slug]/+page.ts
```diff
@@ -31,19 +49,64 @@ export async function load({ url }) {
...
+  const pilot = await getPilot(rigId);
 
   return {
     imageUrl: httpUri,
     badges: [],
     pilot,
   };
 }
+
+const getPilot = async function (rigId, pilotAddress, pilotId): Promise<string> {
```
Comment (joewagner) [inline animation_url/src/routes/[slug]/+page.ts:61]: This holds all the display logic for pilots and trainering and parked.  I believe I got everything covered, but definitely worth looking close at this since there are so many conditionals.

animation_url/src/routes/[slug]/+page.ts
```diff
@@ -31,19 +49,64 @@ export async function load({ url }) {
...
+  if (currentSession.end_time >= currentSession.start_time) {
+    return "";
+  }
+
+  // If we get here it means the rig is piloted. NOTE: this will only show eth mainnet Pilots
+  if (currentSession.pilot_contract && currentSession.pilot_id) {
+    const pilotToken = await zdk.token({
+      token: {
+        address: currentSession.pilot_contract, // "0x6c9343ca5c2ef3a35a83438344bb3cbe3c249f65",
+        tokenId: currentSession.pilot_id.toString(), // "903",
```
Comment (joewagner) [inline animation_url/src/routes/[slug]/+page.ts:102]: You can uncomment the values here if you want to testout what the pilots look like during development

animation_url/src/routes/[slug]/+page.ts
Comment (sanderpick) [reply animation_url/src/routes/[slug]/+page.ts:61]: I think this can be simplified to just do one query right? similar to what i posted in discord?

Comment (sanderpick) [reply animation_url/src/routes/[slug]/+page.ts:61]: ie, we don't care about any sessions where end_time is NOT null

Comment (joewagner) [reply animation_url/src/routes/[slug]/+page.ts:61]: I'll update now...

Comment (joewagner) [reply animation_url/src/routes/[slug]/+page.ts:61]: @sanderpick Sorry again about the confusion here.  I updated things.  There's only one query now, and the logic assumes that a given rig can only ever have one row where end_time is null.  Let me know if this looks right.

sanderpick (approved):
No worries! Looks perfect 👍

PR #463, opened 2022-12-06
Custom pilots UI

datadanne:
I got into a bit of a time-crunch with this so some of the code is pretty low quality and could use some refactoring, but it should work. 

### Some notes:

- All NFT APIs have some sort of downside, I looked at all of them and ended up using Alchemy here but the solution isn't ideal. The reason I had to switch is that Zora's search endpoint was really bad, it wouldn't find common big collections like "Chain Runners".
- The alchemy API doesn't support filtering out ERC1155's or a certain contract address from any of their endpoints, so the "pick your pilot"-grid will include ERC1155s and Rigs but they'll be greyed out with a 🚫 icon over them. NFTPort supports filtering out ERC1155s but their search seemed really slow, and their website was broken and I couldn't access the billing portal to see what their rate limits/pricing looks like so I didn't want to risk switching to them last minute. Will look into using NFTPort or maybe a combo of Alchemy and NFTPort in the future
- Alchemy doesn't host images that are stored on chain using `data:`  uris, so I had to update the components that render NFTs to support both an imageUrl and data uris.
- Filtering owned NFTs by collection doesn't work on testnets, Alchemy's search API doesn't support that at all. Will look into if there are any alternatives for that but their search was the best I could find
- I didn't have time to finish bulk piloting from the Dashboard page, most of the code is there and the modal supports bulk piltoing but I need to add some styling/polish to the 'carousel' in the modal where you rotate between the Rigs you want to Pilot + I need add some filtering to the Dashboard 'inventory' component to make it easier to select the Rigs that are available to Pilot. Will work on that next
- The piloting flow does a check if the pilot is used by another Rig before the user can submit the tx, that warning could use some layout improvements and like show the image of the pilot that is in use etc

### Questions
Any thoughts about the copy here @sanderpick?
image

### Random screenshots
Pick pilot grid that shows how ERC1155s are rendered:
image

Filter pilot grid by collection search
image

Submit pilot tx step without warning text
image

Activity feed showing the collection name of the pilot (nvm that I'm using a rig inside a rig here, it is a Rig from another test contract)
image


datadanne:
Pushed a few small fixes today, should be good to go!

garage/src/hooks/useNFTs.ts
```diff
@@ -1,40 +1,62 @@
...
   return {
-    contract: collectionAddress,
+    type: contract.tokenType,
+    contract: contract.address,
     tokenId,
-    name: tokenName || fallbackName,
-    imageUrl,
+    name: title,
+    imageUrl: media.length ? media[0].gateway : undefined,
+    imageSvgData: data.rawMetadata?.svg_image_data,
```
sanderpick, line 55: 👍 . playing with something similar in animation_url, I noticed that one of my NFTs has a video for the `image` metadata (no real static image provided). i'm not sure how common that is... seems like they should have used `animation_url` for that, but looking into adding a video element to handle mp4 and webm files.

sanderpick (approved):
All looks good to me. I have played with it a good bit as well. 

Lots of work here. Nice job! 🥇 

The warning copy for duplicate pilots looks great.

PR #472, opened 2022-12-07
Let token owner transfer an in-flight Rig (still block approved address / operator)

dtbuchholz:
For reference, see the [Moonbirds contract's](https://etherscan.io/token/0x23581767a106ae21c074b2276d25e5c3e136a68b#code) `safeTransferWhileNesting`. These changes, essentially, do the same thing but include an update to the current session in the `pilot_sessions` table with the token's new `owner` value.

## Summary

Token transfers are blocked with the intention that in-flight Rigs should not be sold. But, transfers between an owner's accounts should be possible. These changes enable a token owner to transfer between accounts using `safeTransferWhileFlying` but continue to block transfers initiated by an approved address or operator.
    
`TablelandRigs.sol`
- Create a new variable `_allowTransferWhileFlying` that stores a `0` (default) or `1`, where `1` is temporarily set to allow for `safeTransferWhileFlying` to occur.
- `safeTransferWhileFlying` *only* allows the token's owner to transfer to another wallet (not approved addresses nor operators) and maintains the current pilot session, without disrupting flight time. It updates the `owner` value in the table using the pilot contract's `updateSessionOwner` method.
- Slight change to `_beforeTokenTransfers` to enable this functionality and allow a transfer if `_allowTransferWhileFlying == 1`.
    
`TablelandRigPilots.sol`
- `updateSessionOwner` updates the `pilot_sessions` table with a new `owner` value; the Rigs contract's `safeTransferWhileFlying` makes the proper checks on token ownership.
    
`ITablelandRigs.sol` & `ITablelandRigPilots.sol`
- Make the proper interface changes related to the new methods noted above.

Tests are also updated to reflect these changes.


ethereum/contracts/TablelandRigs.sol
```diff
@@ -524,6 +527,26 @@ contract TablelandRigs is
...
+    function safeTransferWhileFlying(
+        address from,
+        address to,
+        uint256 tokenId
+    ) external {
+        // Verify `msg.sender` is the token owner (prevent transfers by approved address or operator)
+        if (ownerOf(tokenId) != _msgSenderERC721A())
+            revert ITablelandRigPilots.Unauthorized();
+        // Temporaritly set the transfer flag to allow a transfer to occur
+        _allowTransferWhileFlying = 1;
```
dtbuchholz, line 542: Instead of values `0` and `1` for `_allowTransferWhileFlying`, the Moonbirds contract uses `1` and `2`. I can't think of why, unless it was just preference...

The one thing to note is that `_allowTransferWhileFlying` isn't initialized with any value, so it defaults to `0`. I don't think this has any implications, besides, _maybe_ a little extra gas (like 17 gas) upon each `safeTransferWhileFlying` call.

dtbuchholz: Yeah, I don't get it...and using a `bool` actually decreases gas a bit.

sanderpick: May as well do that then?

dtbuchholz: Definitely -- added it in the latest!

PR #475, opened 2022-12-08
Better media parsing

sanderpick:
The alchemy NFT API responses seem to change based on some indexing. eg, a token with GIF media I was testing resulted in a response that had a gateway URI w/ an extension direct from the contract metadata. Rechecking this morning, looks like Alchemy indexed it and included the `format` param and a gateway link without an extension. That broke my parsing, resulting in showing the question mark image.

This changes the logic to use `format` if available and fallback to parsing the URI to determine if the media is an image or video. 

Also switch to using the indexed thumbnail if available.

### DIFF
--- a/animation_url/src/routes/[slug]/+page.svelte
+++ b/animation_url/src/routes/[slug]/+page.svelte
@@ -54,12 +54,14 @@
 
   .container {
     position: relative;
+    width: 100vh;
+    height: 100vh;
+    margin: 0 auto;
   }
 
   .rig-image {
     display: block;
-    width: 100vw;
-    height: 100vw;
+    width: 100%;
   }
 
   .badges-container {
--- a/animation_url/src/routes/[slug]/+page.svelte
+++ b/animation_url/src/routes/[slug]/+page.svelte
@@ -15,13 +15,15 @@
     {/each}
   </div>
   {#if data.pilot}
-    {#if data.pilot.type === "image"}
-      <img class="pilot" src={data.pilot.uri} alt="pilot" />
-    {:else}
-      <video class="pilot" playsinline autoplay muted loop>
-        <source src={data.pilot.uri}>
-      </video>
-    {/if}
+    <div class="pilot">
+      {#if data.pilot.type === "image"}
+        <img src={data.pilot.uri} alt="pilot" />
+      {:else}
+        <video playsinline autoplay muted loop>
+          <source src={data.pilot.uri}>
+        </video>
+      {/if}
+    </div>
     <span class="warning">Status: In-flight (not transferable).</span>
   {/if}
 </div>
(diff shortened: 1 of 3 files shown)


animation_url/src/routes/[slug]/+page.svelte
```diff
@@ -86,6 +86,7 @@
     right: 0;
     max-height: 37%;
     max-width: 25%;
+    min-width: 25%;
```
Comment (sanderpick) [inline animation_url/src/routes/[slug]/+page.svelte:89]: @datadanne the garage pilot display might need something like this that allows for non-square media. I believe it assumes images are square?

Comment (datadanne) [reply animation_url/src/routes/[slug]/+page.svelte:92]: Ah, I was going to ask you how you handle this in this app yesterday. The garage has a fixed square of 25% x 25% for the pilot and displays the image inside that with `object-fit: contain`. So if you use an NFT that isn't square there will just be dark bg color padding so that the pilot box is still a square

Comment (sanderpick) [reply animation_url/src/routes/[slug]/+page.svelte]: nice, changed to do the same here.

animation_url/src/routes/[slug]/+page.ts
```diff
@@ -77,12 +90,29 @@ const getPilot = async function (rigId: string): Promise<Pilot | undefined> {
...
+      return {
+        uri: pilotToken.rawMetadata?.image_data,
+        type: MediaType.image,
+      };
+    } else if (pilotToken.rawMetadata?.svg_image_data) {
+      return {
+        uri: pilotToken.rawMetadata?.svg_image_data,
+        type: MediaType.image,
+      };
     } else {
```
Comment (sanderpick) [inline animation_url/src/routes/[slug]/+page.ts:116]: fallback handling for `image_data` and the lesser used `svg_image_data`

animation_url/src/routes/[slug]/+page.ts
```diff
@@ -92,24 +122,26 @@ const getPilot = async function (rigId: string): Promise<Pilot | undefined> {
...
+  if (!uri) {
+    return undefined;
+  }
+  if (uri.startsWith("data:image/svg+xml;utf8,")) {
+    const parts = uri.split("data:image/svg+xml;utf8,");
+    return "data:image/svg+xml;base64," + btoa(parts[1]);
+  } else {
+    return uri;
+  }
+};
```
Comment (sanderpick) [inline animation_url/src/routes/[slug]/+page.ts:147]: some big svgs, eg, cryptopunks, are not base64 encoded and for whatever reason that combination makes them not render in an image tag. however, they will render when base64 encoded.

ethereum/scripts/parkRigs.ts
```diff
@@ -0,0 +1,33 @@
...
+    /* rigs IDs to park here */
+  ]);
+  const receipt = await tx.wait();
+  console.log(`parked rigs with txn '${receipt.transactionHash}'`);
+}
+
+main().catch((error) => {
+  console.error(error);
+  process.exitCode = 1;
+});
```
sanderpick, line 33: snuck this in here for parking rigs.

PR #480, opened 2022-12-09
Garage: update Dashboard Activity module rows to link to /rigs/:id

datadanne:
Makes the rig image and #id text in the Activity module on the Dashboard clickable links to /rigs/:id


sanderpick (approved):
👍

PR #481, opened 2022-12-09
Garage: batch piloting v1

datadanne:
Not the most amazing user experience but it gets the job done! I think we should add a way to filter Rigs on the dashboard by status later to make it easier to see which rigs you can select. 

In the screenshots below I'm piloting three rigs, so the pilot I'm choosing in step 1 is disabled in 2 etc so that you can't accidentally select the same pilot again.
image
image
image


dtbuchholz:
Note: should probably have this bug fix deployed first: https://github.com/tablelandnetwork/rigs/pull/482

sanderpick (approved):
Lgtm! 💪

PR #483, opened 2022-12-10
Animation url: Simplify pilot uri to only use images from the alchemy gateway or raw svg data

datadanne:
This simplifies the image url calculation to work the same way it does in the garage app. This also fixes the issue reported on discord where moonbirds currently don't work, they use a regular centralized http app to return the metadata without a file extension in the url (like `https://live-metadata....app/1231`). I don't know if this will work with birds that are rendered on-chain though, the alchemy api still returns the old centralized metadata for all birds including those that have moved to on-chain. I created a ticket in the alchemy discord and asked why that is and how it will work once it is fixed, will follow up with the necessary changes once I hear back from them

### DIFF
--- a/animation_url/src/routes/[slug]/+page.ts
+++ b/animation_url/src/routes/[slug]/+page.ts
@@ -47,26 +47,7 @@ export async function load({ url }) {
 
 type Pilot = {
   uri?: string;
-  type?: MediaType;
 };
-
-enum MediaType {
-  image = "image",
-  video = "video",
-}
-
-const mediaTypes = new Map([
-  ["jpg", MediaType.image],
-  ["jpeg", MediaType.image],
-  ["png", MediaType.image],
-  ["svg", MediaType.image],
-  ["gif", MediaType.image],
-  ["mp4", MediaType.video],
-  ["mov", MediaType.video],
-  ["ogv", MediaType.video],
-  ["webm", MediaType.video],
-  ["3gp", MediaType.video],
-]);
 
 const getPilot = async function (rigId: string): Promise<Pilot | undefined> {
   // Get the sessions where end_time is null, there should only ever be one of these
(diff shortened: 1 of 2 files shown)


animation_url/src/routes/[slug]/+page.ts
```diff
@@ -90,58 +71,13 @@ const getPilot = async function (rigId: string): Promise<Pilot | undefined> {
...
-    } else {
-      return { uri: unknown, type: MediaType.image };
-    }
-  }
 
-  // else show trainer
-  return { uri: trainer, type: MediaType.image };
-};
+    const imageUrl = pilotToken?.media[0]?.gateway;
+    const rawSvg = pilotToken?.rawMetadata?.svg_image_data;
```
Comment (sanderpick) [inline animation_url/src/routes/[slug]/+page.ts:76]: happy to get rig of all the ugliness here, but...

- i have seen cases where `gateway` is empty, but `raw` has a value
- worth using thumbnail? some image are many mbs
- `image_data` can be present instead of `svg_image_data` according to OS metadata standards 

why not,

```
const imageUrl = pilotToken?.media[0]?.thumbnail || pilotToken?.media[0]?.gateway || pilotToken?.media[0]?.raw;
const rawSvg = pilotToken?.rawMetadata?.svg_image_data || pilotToken?.rawMetadata?.image_data;
```

`encodeSVG` was used to handle giant SVGs (like crypto punks). for some reason, they do not render unless they are base64 encoded... seems to be related to being giant.

Comment (datadanne) [reply animation_url/src/routes/[slug]/+page.ts:76]: Ah I have never seen the case where gateway is empty and raw has a value. We should totally figure out what gives us the most compatibility, I just threw this together quick this morning because there are currently a few pilots that are broken in prod like moonbirds (because they use `https://gate.way/id` urls without a file ending), and since this is what we currently have in the garage I just know that that works for moonbirds at least

Comment (sanderpick) [reply animation_url/src/routes/[slug]/+page.ts:76]: sounds good. just sat down for a few min here. let's just merge this and I can follow up with a little tweak

sanderpick (approved):

PR #484, opened 2022-12-10
more animation url tweaks

sanderpick:
- use thumbnail when possible
- include check for `image_data`
- handle large svgs


datadanne (approved):
lgtm!

PR #500, opened 2022-12-19
Garage: add account and pilot stats to Dashboard

datadanne:
Adds tabs to the Stats section on the dashboard, and adds two new tabs with account stats & pilot stats:
image
image

This isn't something we have on the current roadmap but it was the top suggestion from the community so it felt right to get that in now :)

@dtbuchholz @sanderpick any suggestions on what to use as labels for the two tables?

### DIFF
--- a/garage/src/App.tsx
+++ b/garage/src/App.tsx
@@ -97,6 +97,16 @@ const theme = extendTheme(
           orbitron: { fontFamily: "'Orbitron', sans-serif", fontWeight: 900 },
         },
       },
+      Tabs: {
+        variants: {
+          line: {
+            tab: {
+              color: "primary",
+              _selected: { color: "primary" },
+            },
+          },
+        },
+      },
       Table: {
         variants: {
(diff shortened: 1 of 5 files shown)


datadanne:
> damn, very cool. by labels do you mean what you currently are calling "you" and "pilots"? if so, i think those make perfect sense.

Ah I meant "Top collections" and "Most FT earned", for some reason every header/label I tried felt wrong when I wrote those haha

sanderpick:
Makes sense. Tricky labeling, but I think what you have is fine.

I think ideally there is just one table with columns for num. pilots and FT. Click each column header to sort by it. That works avoid the need for labels.

datadanne:
> Makes sense. Tricky labeling, but I think what you have is fine.

Ok cool!

> I think ideally there is just one table with columns for num. pilots and FT. Click each column header to sort by it. That works avoid the need for labels.

Yeah that could make sense. I'll try that later!

PR #521, opened 2023-01-03
Garage: Add owner profiles

datadanne:
What do you think about this layout? Anything that is missing that we should add?

image

I have a few things left to fix:
- The header should contain the total FT earned by the wallet
- We should add a small "Owned by: <wallet>" text to the /rigs/:id page so that there is a natural place to link from to /owner/:wallet from rig details
- We should translate the wallet address to an ENS address if available but the ens name translation seems to broken in the version of wagmi that we are using, I'll try adding that again after I have upgraded that

but I think we can merge this PR as it is and do that in a follow up change

Closes https://github.com/orgs/tablelandnetwork/projects/56


datadanne:
Added the total FT to the header, good to go now

dtbuchholz:
lgtm. one thing -- it took me a bit to find where to access `/owner/<address>`. maybe the address in the top right should at least have a similar `text-decoration: underline` on it make it a button, or some way to emphasize the feature.

dtbuchholz (approved):

datadanne:
> lgtm. one thing -- it took me a bit to find where to access /owner/<address>. maybe the address in the top right should at least have a similar text-decoration: underline on hover, make it a button, or some way to emphasize the feature.

Yeah it is kind of hidden right now. I think adding a "This Rig is owned by: <wallet>" blurb to the rigs detail page and have that link to the owners profile will help make it much easier to discover, but the address should definitely have a underline on hover or something as well

Issue #414, opened 2022-11-18
Implement ability to send / transfer rigs that are parked


sanderpick:
After playing with the "select -> do action" UI, I realized it would be nice to add the ability to transfer / bulk transfer rigs. Nice thing about doing it in the UI, is that we could show a greyed out "transfer" button / icon on the rig page if in-flight to drive home the fact that it's not transferable.

PR #539, opened 2023-01-11
Garage: add support for transferring rigs

datadanne:
This PR adds support for transferring a single rig from the /rigs/:id page. It supports both parked and in-flight rigs (using the `safeTransferWhileFlying` method).

This implementation is a bit different than what the issue described. My reasoning:
- I think we should have a way for holders to transfer rigs that are in-flight. If a holder believes that their wallet is compromised or something like that we shouldn't block them from transferring their Rigs in a safe way. The soft-staking kind of works as a protection against theft since a hacker wouldn't be able to transfer assets using a typical bulk transfer contract or by calling `safeTransferFrom` or by sell it via an already approved marketplace, but that only works as long as the owner can transfer it without parking. Moonbirds does this too
- There is no way to bulk transfer NFTs without using a separate contract, the ERC721 interface only supports transferring one token at a time. Marketplaces, MEV bots etc all do this by using custom contracts, we could do that too and add a method to bulk transfer to the contract if we believe that is a common use case? 
- We could add a button to the dashboard that lets you transfer a single rig from there as well if you've only selected one rig, I tried that but it didn't feel very intuitive to use since the other button supports bulk actions   

image
image
image

Closes https://github.com/tablelandnetwork/rigs/issues/414 ish

### DIFF
--- a/garage/src/components/TransferRigModal.tsx
+++ b/garage/src/components/TransferRigModal.tsx
@@ -1,0 +1,154 @@
+import React, { useCallback, useEffect, useMemo, useState } from "react";
+import {
+  Button,
+  FormControl,
+  FormLabel,
+  Input,
+  Modal,
+  ModalOverlay,
+  ModalContent,
+  ModalHeader,
+  ModalFooter,
+  ModalBody,
+  ModalCloseButton,
+  Text,
+} from "@chakra-ui/react";
+import { WarningTwoIcon } from "@chakra-ui/icons";
+import {
+  useAccount,
+  useContractWrite,
+  usePrepareContractWrite,
+  useWaitForTransaction,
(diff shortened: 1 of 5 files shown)


dtbuchholz:
If owners want it, then I'm all for adding a bulk method. I just don't know how common it'll be used. 

I suppose if the "most often used scenario" is when a wallet is compromised, it'd be nice to be able to batch that transfer. But, the owner would have to be aware of the method via direct contract or The Garage UI. Just making that point re: is it worthwhile.

sanderpick (approved):
Looks great! +1 on skipping some kind of bulk transfer implementation. If people really want to do that, they could use the opensea UI, then revoke approval.

PR #548, opened 2023-01-16
Garage: Improve topbar search input and rig details on mobile

datadanne:
- Moves the topbar search input into a modal on mobile, works much better on small screens and especially if the user doesn't have a wallet connected since that button is much wider than the exit button
- Truncates the `Owned by: <wallet>` address on mobile screens to avoid a horizontal scroll on that view

image
image
image


sanderpick (approved):
nice tweak 👍

PR #560, opened 2023-01-23
Garage: update tableland sdk to v4

datadanne:
Unblocked! 🎉 

~UPDATE: Blocked by https://github.com/tablelandnetwork/js-tableland/issues/390~

~UPDATE: Blocked by https://github.com/tablelandnetwork/js-tableland/issues/383~

~456fb183d25ca4103d2a8d5b95df89ba008ecb80 is a temporary change that we won't need after we've published a new version of @tableland/rigs~

~This is a draft PR because it is currently blocked by another PR, an issue with the rigs-package not being compatible with the v4 sdk and the sdk v4 sqlparser wasm-file not being served correctly in vite dev-setups.~

~Blocked by https://github.com/tablelandnetwork/rigs/pull/542~
~Blocked by https://github.com/tablelandnetwork/rigs/issues/559~

Closes https://github.com/tablelandnetwork/rigs/issues/558

Issue #559, opened 2023-01-23
Publish a new version of @tableland/rigs that uses the v4 sdk


datadanne:
The current version of @tableland/rigs can't be used with @tableland/sdk v4 because of
```
node_modules/@tableland/rigs/deployments.ts(54,5): error TS2322: Type '"polygon-mumbai"' is not assignable to type 'keyof TablelandNetworkConfig'.
```

I also noticed that @tableland/rigs doesn't specify the tableland sdk as a dependency which it probably should?

joewagner:
Do we need to capture the work to update the animation_url?  Or is that part of one of the above draft PRs?

asutula:
There's some discussion happening in Discord about that. Not clear what the path is, but yes, we should capture it as an issue and eventual PR.

asutula:
@joewagner pls ignore my above comment. I mis understood what you were talking about. Yes, we should capture an issue to update the `animation_url` app in this repo.

asutula:
updated this issue with a checklist for the different things that need the sdk update https://github.com/tablelandnetwork/rigs/issues/558

PR #561, opened 2023-01-23
Garage: Rigs gallery

datadanne:
Ready to go! 🚀 

~A WIP gallery page. Would love some overall feedback on how this feels to use and ideas on features/data that is missing
etc. I've listed some known issues and ideas on what we could add below~

~UPDATE: deployments should work now!~
~NOTE:~
~The automated deployment doesn't work for this branch since it is based on the SDK v4 update PR (https://github.com/tablelandnetwork/rigs/pull/560), and that PR is broken because the `@tableland/rigs` package doesn't work with the V4 SDK (issue here: https://github.com/tablelandnetwork/rigs/issues/559). You can run it locally though if you want to play with it~

image
image

###
image
image
image

### TODO
- [x] Figure out some type of main navigation in the top bar that works on both desktop and mobile since we'll have two "top level" pages now: the dashboard and the gallery
- [x] Mobile layout, we should probably do something similar to what I did on the rigs-explorer site where you toggle the filters list like a popup on mobile? 
- [x] Improve performance when enabling/disabling a filter by clicking on the checkbox in the filters module
- [x] Overall layout cleanup: remove the random Gallery heading and make sure everything looks & feels like the rest of the garage with consistent spacing, headings etc

### OTHER IDEAS
- [ ] Combine background filters so that we only have "Dark Alpine" instead of "Dark Alpine 1", "Dark Alpine 2", "Dark Alpine 3" etc
- [ ] Bind the current filters to query string parameters so that links with filters are shareable? This is usually kind of annoying to implement so that it doesn't mess up browser navigation etc but might be worth it if we think the gallery will be used a lot
- [ ] If we make filters query parameters, maybe add links from the rig details page attributes breakdown so that you can find other rigs with the same attribute? Like if you are looking at a Titan Rockbiter-original, make Rockbiter link to the gallery page with a filter "Name: Rockbiter" added

### CLEAN UP STUFF
- [x] Break out the footer with OS & Twitter link into a shared component (copied from the dashboard right now)
- [x] Break out gallery components into their own files, general code clean up
- [ ] Look into if we can query for all attributes on page load, I ran some queries manually and dumped into a .json-file to avoid the page feeling slow to load but since the tables are pretty small the queries should be fast so we should maybe try that. We will also have to look into validator rate-limits soon, if you click around fast enough some requests gets rejected with a 429

Blocked by https://github.com/tablelandnetwork/rigs/pull/560
Closes https://github.com/tablelandnetwork/rigs/issues/408


andrewxhill:
- Any thoughts about a "Buy Now" filter that links to OS or others if it's listed? I know OS must be gettable. 
- Sort by FT could be a nice link. 

Looking solid.

datadanne:
> Any thoughts about a "Buy Now" filter that links to OS or others if it's listed? I know OS must be gettable.

Yeah I think that would be a great thing to add! I think the simplest (and maybe only?) way to implement that would be to have the client fetch all rig listings, and if that filter is enabled append a `WHERE rig_id IN (...listedRigs)` clause to all queries. Maybe not the most scaleable solution but that should be fine for a small collection like Rigs I think
 
> Sort by FT could be a nice link.

Yeah that could be cool! 

----

I realized now that I think we should add filters for dynamic stats/attributes as well, like
- `Is in-flight               [ ]`
- `Is trained                 [ ]`
- `Has used non-trainer pilot [ ]`
- `Has earned at least _______ FT`

andrewxhill:
oh, this one wouldn't need to be new interface. but the ability to click on a collection name from the leaderboard and it would take you into the gallery page with a filter to just those rigs using that collection would be spicy. 

Screen Shot 2023-01-24 at 8 14 55 AM

stretch goal. would it be overkill in that case to render the animation_url for every tile?

datadanne:
> oh, this one wouldn't need to be new interface. but the ability to click on a collection name from the leaderboard and it would take you into the gallery page with a filter to just those rigs using that collection would be spicy.
>
> stretch goal. would it be overkill in that case to render the animation_url for every tile?

ah this is a great idea, I like it!

we can show the animation_url for tiles in the gallery, it is a little bit annoying though since we have to query the alchemy api for the nft images after the rig query has loaded so there will always be a noticeable delay between the rig image loading and the pilot image loading, but definitely doable. I think we'd have to be consistent and always show it in the gallery if we show it when a collection filter is enabled

dtbuchholz (approved):

datadanne:
Here we go!

PR #562, opened 2023-01-23
Update @tableland/rigs to the latest sdk

### LINKED ISSUE #559: Publish a new version of @tableland/rigs that uses the v4 sdk

asutula:
Pretty straight forward I think, but blocked now by some upstream work on getting the sdk to properly provide a cjs distribution (https://github.com/tablelandnetwork/wasm-sqlparser/pull/91, which will be pulled into the sdk). This is required because hardhat projects like this are cjs. It's not clear to me if they can be esm or not. I tried and quickly ran into problems with hardhat loading the hardhat config file. 

Will leave this as draft for now for some early eyes.

Closes #559


datadanne (changes requested):
Nice, the code changes lgtm!

We need to update the import in `ethereum/deployments.ts` as well, `@tableland/sdk` isn't exporting `ChainName`:
```
deployments.ts(1,10): error TS2305: Module '"@tableland/sdk"' has no exported member 'ChainName'.
```

asutula:
Good catch @datadanne, fixed.

carsonfarmer (approved):

datadanne (approved):

asutula:
Version 0.1.3 is published to npm @datadanne

Issue #558, opened 2023-01-23
Update tableland sdk to v4


datadanne:
- [x] ethereum
- [ ] garage
- [ ] animation_url

PR #567, opened 2023-01-25
Update the marketplace animation_url app to use newest sdk version

joewagner:
Part of #558

A few things to note:
- Looks like we will be able to get the new wasm based sdk working with vite by setting vite config to not optimize the `@tableland/sqlparser` package.


animation_url/src/routes/[slug]/+page.ts
```diff
@@ -27,13 +36,15 @@ export async function load({ url }) {
...
-    output: "objects",
-    extract: true,
-    unwrap: true,
-  });
-  const httpUri = ipfsGatewayUri + metadata.image.slice(7);
+  ) FROM ${attributes} JOIN ${lookups} WHERE rig_id=${rigId} GROUP BY rig_id;`;
+  const metadata = await db.prepare(stm).all();
+  const image = metadata.results[0][
+    "json_object('image', 'ipfs://' || renders_cid || '/' || rig_id || '/' || image_medium_name)"
+  ].image;
```
Comment (datadanne) [inline animation_url/src/routes/[slug]/+page.ts:43]: I think the query & the code here could be simplified to something like
```ts
const stm = `SELECT renders_cid, image_medium_name FROM ${lookups}`

const { renders_cid, image_medium_name } = await db.preapre(stm).first<{ renders_cid: string, image_medium_name: string }>();
const image = `ipfs://${renders_cid}/${rig_id}/${image_medium_name}`;
```

animation_url/vite.config.ts
```diff
@@ -14,6 +14,11 @@ const config: UserConfig = {
       allow: [".."],
     },
   },
+  optimizeDeps: {
+    exclude: [
+      "@tableland/sqlparser"
```
joewagner, line 19: This is a fix for the error:
```
Not found: /node_modules/.vite/deps/main.wasm
Error: Not found: /node_modules/.vite/deps/main.wasm
```
It would be nice to fix this elsewhere, but I can't find any other solutions.

joewagner:
All the upstream changes have been published and this is ready for review.

@datadanne Thanks for those fixes! I tried to include them here. Let me know if that all looks correct now.  Also, I'm not familiar with the test rigs collection.  I looked at the first half dozen of them and it seems correct afaict.

datadanne (approved):
Changes LGTM!

Issue #395, opened 2022-11-16
Implement a pilot page


sanderpick:
This would be similar to rig page, but flipped (show earned FT, pilot image, traits, rig history, activity)

PR #629, opened 2023-02-16
Garage: Add Pilot page

datadanne:
Adds a detail page for pilots.
image

Links from RigDetails Pilot module
image

Closes https://github.com/tablelandnetwork/rigs/issues/395

### DIFF
--- a/garage/src/routes.tsx
+++ b/garage/src/routes.tsx
@@ -2,6 +2,7 @@ import React from "react";
 import { Enter } from "./pages/Enter";
 import { Dashboard } from "./pages/Dashboard";
 import { RigDetails } from "./pages/RigDetails";
+import { PilotDetails } from "./pages/PilotDetails";
 import { OwnerDetails } from "./pages/OwnerDetails";
 import { Gallery } from "./pages/Gallery";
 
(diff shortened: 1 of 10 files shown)


garage/src/routes.tsx
```diff
@@ -25,5 +26,11 @@ export const routes = () => {
       requiresWalletConnection: false,
       element: <OwnerDetails />,
     },
+    {
+      key: "PILOT_DETAILS",
+      path: "/pilot/:collection/:id",
```
sanderpick, line 31: how about plural instead? ie, `"/pilots/:collection/:id"`

sanderpick (commented):
left a small comment.

as a side note, I think it would look nice to add a little extra space below the top nav bar to separate it from the main content (like 2x what is there now), at least on larger screens. the pilot page kinda drives that home since they will be so short in height

garage/src/routes.tsx
datadanne: Sure!

datadanne:
> as a side note, I think it would look nice to add a little extra space below the top nav bar to separate it from the main content (like 2x what is there now), at least on larger screens. the pilot page kinda drives that home since they will be so short in height

Yeah that sounds like a good idea, I'll play with that!

datadanne:
@sanderpick I doubled the padding between the topbar and content on all pages and it looks good imo! Pushed some new commits that changes the top padding + changes the route name + adds a missing link from the collector profile pilots module to the new pilot page

sanderpick (approved):
lovely!

datadanne:
@sanderpick could you re-review this one? I had to rebase it on `main` to get the tests to pass

sanderpick (approved):
👍

Issue #646, opened 2023-02-20
Fix park pilot modal copy 


datadanne:
RIght now it always shows a "Training isn't complete" warning, but it shouldn't if the Rig has completed training

PR #652, opened 2023-02-21
Garage: Fix park modal to check if training is complete before warning about it

datadanne:
Fixes the park modal so that it doesn't always warn about training not being complete. The modal looks kind of empty without that text, if you want to add some other text if the rig is trained feel free to @sanderpick 

image

Closes https://github.com/tablelandnetwork/rigs/issues/646


sanderpick (approved):
Nice catch. I think we can update this text when tackling the docs work.

PR #654, opened 2023-02-22
Garage: Add raw metadata query link to Rig details page

datadanne:
Adds a Tableland icon on the Rig details page that links to the raw metadata query. The query is fetched from the contracts tokenURI.

Also changes how we use svgs so that we can set the fill color via css so that we don't have to create multiple versions of the same svg

image

Closes https://github.com/tablelandnetwork/rigs/issues/550


sanderpick (approved):
Slick!

Issue #632, opened 2023-02-16
[RIG-23] Add “in-flight” emoji to rig names


sanderpick:
null

<sub>[RIG-23](https://linear.app/tableland/issue/RIG-23/add-in-flight-emoji-to-rig-names)</sub>

PR #665, opened 2023-02-27
Rigs: update metadata query to include a ✈️-emoji in the name property for in-flight rigs

datadanne:
The name field part of the query:
```sql
case 
  when exists(
    select * 
    from ${pilotsTable} 
    where rig_id = ${attributesTable}.rig_id and end_time is null
  ) then 'Rig #'||rig_id||' ✈️' 
  else 'Rig #'||rig_id 
end
```

We could also include the string in-flight instead of an emoji if we think that is clearer, like `Rig #53 (in-flight)`. What do you think?

Closes https://github.com/tablelandnetwork/rigs/issues/632


sanderpick (dismissed):
🥳 I like calling them out with an emoji... looks a bit more playful.

datadanne:
> 🥳 I like calling them out with an emoji... looks a bit more playful.

Yeah I agree! I think you will have to deploy this since you can access the deployer keys? Should I merge it or leave it to you to do that when you deploy?

I'm also wondering if we should wait with deploying this until we've implemented support for on-chain metadata refreshes (which shouldn't take long I think, have we planned when we are going to implement that yet @dtbuchholz?)? It might be confusing instead of helpful if we've implemented this but don't autorefresh metadata since parked rigs could have the ✈️  emoji in their name then (if metadata was refreshed while the rig was in-flight and it got parked after)

sanderpick:
Good point... agreed we should wait until the metadata refresh lands!

dtbuchholz:
@datadanne yeah, i can get to that after ETHDenver this week if that works.

dtbuchholz:
Reference: https://github.com/tablelandnetwork/rigs/pull/688

datadanne:
Just rebased this so it should be ready to go once we merge #688

datadanne:
@sanderpick I did some final testing yesterday and noticed that I had to make a small change, can you re-review? 🙏

sanderpick (approved):

PR #669, opened 2023-02-28
Garage: Show FT under each rig on dashboard

datadanne:
Updates the dashboard to use a ✈️ to indicate if a rig is in-flight, and replaces the in-flight text with total FT earned:
image
instead of
image

I tried loading the dashboard for the largest holder (`0x6CaB818DAaF4040476E2489A7ED29Ca685aD0623`) and the query still loads fast enough!

Closes https://github.com/tablelandnetwork/rigs/issues/399


sanderpick (approved):
sweet!

PR #670, opened 2023-02-28
Use UNION to extend metadata attributes

asutula:
When doing some recent work to extend Rigs metadata with Filecoin deal information, I was finding the existing metadata sql query very difficult to further extend since it had already been extended to support Garage status by using the `json_insert` SQLite function. Felt like there must be a better way that supports further metadata extensions without further complicating our use of SQLite JSON functions. Spent a while figuring it out, and landed on this. This query design fits nicely with the to-be-created Filecoin deals table and is overall just easier to read and understand.

### DIFF
--- a/ethereum/helpers/uris.ts
+++ b/ethereum/helpers/uris.ts
@@ -1,72 +1,90 @@
-export function getURITemplate(
+import { init, __wasm } from "@tableland/sqlparser";
+
+export async function normalize(sql: string) {
+  if (__wasm == null) {
+    await init();
+  }
+  return (await globalThis.sqlparser.normalize(sql)).statements[0];
+}
+
+export async function getURITemplate(
   tablelandHost: string,
   attributesTable: string,
   lookupsTable: string,
   pilotsTable: string,
   displayAttributes: boolean
-): string[] {
+): Promise<string[]> {
   if (!displayAttributes) {
     const uri =
       tablelandHost +
-      "/query?extract=true&unwrap=true&s=" +
+      "/api/v1/query?extract=true&unwrap=true&statement=" +
       encodeURIComponent(
-        `select json_object(
-          'name','Rig #'||rig_id,
-          'external_url','https://garage.tableland.xyz/rigs/'||rig_id,
-          'image','ipfs://'||renders_cid||'/'||rig_id||'/'||image_full_name,
-          'image_alpha','ipfs://'||renders_cid||'/'||rig_id||'/'||image_full_alpha_name,
-          'image_medium','ipfs://'||renders_cid||'/'||rig_id||'/'||image_medium_name,
-          'image_medium_alpha','ipfs://'||renders_cid||'/'||rig_id||'/'||image_medium_alpha_name,
-          'thumb','ipfs://'||renders_cid||'/'||rig_id||'/'||image_thumb_name,
-          'thumb_alpha','ipfs://'||renders_cid||'/'||rig_id||'/'||image_thumb_alpha_name,
-          'animation_url',animation_base_url||rig_id||'.html',
-          'attributes',json_array(
+        await normalize(
+          `select 
             json_object(
-              'trait_type','status',
-              'value','pre-reveal'
+              'name','Rig #'||rig_id,
+              'external_url','https://garage.tableland.xyz/rigs/'||rig_id,
+              'image','ipfs://'||renders_cid||'/'||rig_id||'/'||image_full_name,
+              'image_alpha','ipfs://'||renders_cid||'/'||rig_id||'/'||image_full_alpha_name,
+              'image_medium','ipfs://'||renders_cid||'/'||rig_id||'/'||image_medium_name,
+              'image_medium_alpha','ipfs://'||renders_cid||'/'||rig_id||'/'||image_medium_alpha_name,
+              'thumb','ipfs://'||renders_cid||'/'||rig_id||'/'||image_thumb_name,
+              'thumb_alpha','ipfs://'||renders_cid||'/'||rig_id||'/'||image_thumb_alpha_name,
+              'animation_url',animation_base_url||rig_id||'.html',
+              'attributes',json_array(
+                json_object(
+                  'trait_type','status',
+                  'value','pre-reveal'
+                )
+              )
             )
-          )
+          from
+            ${attributesTable}
+            join ${lookupsTable}
+          where rig_id=ID
+          group by rig_id;`
         )
-        from ${attributesTable} join ${lookupsTable}
-        where rig_id=ID
-        group by rig_id;`.replace(/(\r\n|\n|\r|\s\s+)/gm, "")
       );
     return uri.split("ID");
   } else {
     const uri =
       tablelandHost +
-      "/query?extract=true&unwrap=true&s=" +
+      "/api/v1/query?extract=true&unwrap=true&statement=" +
       encodeURIComponent(
-        `select json_object(
-          'name','Rig #'||rig_id,
-          'external_url','https://garage.tableland.xyz/rigs/'||rig_id,
-          'image','ipfs://'||renders_cid||'/'||rig_id||'/'||image_full_name,
-          'image_alpha','ipfs://'||renders_cid||'/'||rig_id||'/'||image_full_alpha_name,
-          'image_medium','ipfs://'||renders_cid||'/'||rig_id||'/'||image_medium_name,
-          'image_medium_alpha','ipfs://'||renders_cid||'/'||rig_id||'/'||image_medium_alpha_name,
-          'thumb','ipfs://'||renders_cid||'/'||rig_id||'/'||image_thumb_name,
-          'thumb_alpha','ipfs://'||renders_cid||'/'||rig_id||'/'||image_thumb_alpha_name,
-          'animation_url',animation_base_url||rig_id||'.html',
-          'attributes', json_group_array(
-            json_object('display_type',display_type,'trait_type',trait_type,'value',value)
-          )
+        await normalize(
+          `select
+            json_object(
+              'name','Rig #'||rig_id,
+              'external_url','https://garage.tableland.xyz/rigs/'||rig_id,
+              'image','ipfs://'||renders_cid||'/'||rig_id||'/'||image_full_name,
+              'image_alpha','ipfs://'||renders_cid||'/'||rig_id||'/'||image_full_alpha_name,
+              'image_medium','ipfs://'||renders_cid||'/'||rig_id||'/'||image_medium_name,
+              'image_medium_alpha','ipfs://'||renders_cid||'/'||rig_id||'/'||image_medium_alpha_name,
+              'thumb','ipfs://'||renders_cid||'/'||rig_id||'/'||image_thumb_name,
+              'thumb_alpha','ipfs://'||renders_cid||'/'||rig_id||'/'||image_thumb_alpha_name,
+              'animation_url',animation_base_url||rig_id||'.html',
+              'attributes', json_group_array(
+                json_object('display_type',display_type,'trait_type',trait_type,'value',value)
+              )
+            )
+          from 
+            (
+              select *
+              from ${attributesTable}
+              union
+              select
+                a.rig_id,
+                'string' display_type,
+                'Garage Status' trait_type,
+                case when start_time is null then 'parked' else 'in-flight' end value
+              from
+                ${attributesTable} a
+                left join (select * from ${pilotsTable} where end_time is null) s on a.rig_id = s.rig_id
+            )
+            join ${lookupsTable}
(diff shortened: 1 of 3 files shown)


ethereum/helpers/uris.ts
```diff
@@ -26,10 +26,10 @@ export function getURITemplate(
               'value','pre-reveal'
             )
           )
-        ) from ${attributesTable} join ${lookupsTable} where rig_id=ID group by rig_id;`.replace(
-          /(\r\n|\n|\r|\s\s+)/gm,
-          ""
         )
+        from ${attributesTable} join ${lookupsTable}
+        where rig_id=ID
+        group by rig_id;`.replace(/(\r\n|\n|\r|\s\s+)/gm, "")
```
asutula, line 32: No real change here, just formatting.

ethereum/helpers/uris.ts
```diff
@@ -47,27 +47,26 @@ export function getURITemplate(
...
-            ),
-            '$[#]',
-            json_object(
-              'display_type','string',
-              'trait_type','Garage Status',
-              'value',coalesce(
-                (select coalesce(end_time, 'in-flight') from ${pilotsTable} where rig_id=ID and end_time is null),
-                'parked'
-              )
-            )
```
asutula, line 65: This was the code that was really hard to further extend. So now there is no more use of `json_insert`.

ethereum/helpers/uris.ts
```diff
@@ -47,27 +47,26 @@ export function getURITemplate(
...
+          'attributes', json_group_array(
+            json_object('display_type',display_type,'trait_type',trait_type,'value',value)
```
asutula, line 51: Now we go back to simply using the `json_group_array` aggregate function which now just operates on a single result set created below by `union`ing together many datasets.

ethereum/helpers/uris.ts
```diff
@@ -47,27 +47,26 @@ export function getURITemplate(
...
+          union
+          select
+            a.rig_id,
+            'string' display_type,
+            'Garage Status' trait_type,
+            case when start_time is null then 'parked' else 'in-flight' end value
+          from
+            ${attributesTable} a
+            left join (select * from ${pilotsTable} where end_time is null) s on a.rig_id = s.rig_id
+        )
```
asutula, line 66: This part of the `from` clause in between this set of parentheses has one goal: Select data from any/all tables you want to create metadata from, and combine them into a single virtual dataset by using `union`. So each `union`ed query must return the same columns that work with our json functions further above: `rig_id integer, display_type text, trait_type text, value text`. In this case we're just `union`ing the attributes table with data from the pilot sessions table, but it will be easy to add another `union` with data from any new table, like Filecoin deals, for example.

ethereum/helpers/uris.ts
```diff
@@ -47,27 +47,26 @@ export function getURITemplate(
...
+        join ${lookupsTable}
```
asutula, line 67: This `join` just makes our lookups columns available to us like before. No change here.

ethereum/helpers/uris.ts
```diff
@@ -47,27 +47,26 @@ export function getURITemplate(
...
+        where rig_id=ID
+        group by rig_id;`.replace(/(\r\n|\n|\r|\s\s+)/gm, "")
```
asutula, line 69: `filter` and `group by` remain the same as before.

ethereum/helpers/uris.ts
```diff
@@ -47,27 +47,26 @@ export function getURITemplate(
...
+          from ${attributesTable}
+          union
+          select
+            a.rig_id,
+            'string' display_type,
+            'Garage Status' trait_type,
+            case when start_time is null then 'parked' else 'in-flight' end value
+          from
+            ${attributesTable} a
+            left join (select * from ${pilotsTable} where end_time is null) s on a.rig_id = s.rig_id
```
asutula, line 65: This `left join` with a subquery is a little trick to detect which rigs currently are in flight. The logic is that since we're `left join`ing all rig ids with pilot sessions where `end_time` is null, only in flight rig ids will end up with a non-null `start_time` in the joined data, and parked rigs will have a null `start_time`.

ethereum/helpers/uris.ts
asutula: Hmm, it seems the regex that removes spaces etc is sensitive to some ways of formatting i'm doing here. This is causing a test failure. @sanderpick it seems like we should be free to format this however is easy to read. Maybe we need to improve the regex? Or document what the formatting limitations are here? (pls let me know, it's not clear to me yet).

datadanne: The reason why it breaks a test is that the regexp replaces all occurrences of multiple spaces (`\s\s+`) with `""`, with the new formatting in this PR the regexp ends up removing all spaces between `${lookupsTable}` and `where rig_ID...` since they are split into two lines and have multiple spaces between them, where as before it was just one space. The query right returned right now is actually invalid, `...from table2where rig_ID=...`

IMO the tests aren't really helping much they way they are written now though since we have to copy/paste all query changes to the tests, and the string diff output is hard to read. I think we should:

- replace the regexp used to remove spaces and newlines with:
```js
import { init, __wasm } from "@tableland/sqlparser";

const normalize = async (sql: string) => {
  if (__wasm == null) {
    await init();
  }
  return (await sqlparser.normalize(sql)).statements[0];
};

export async function getURITemplate(
  tablelandHost: string,
  attributesTable: string,
  lookupsTable: string,
  pilotsTable: string,
  displayAttributes: boolean
): Promise<string[]> {
  if (!displayAttributes) {
    const uri =
      tablelandHost +
      "/query?extract=true&unwrap=true&s=" +
      encodeURIComponent(
        await normalize(`select json_object(
          ...
```

- update the tests to:
1) parse the `s` query parameter and check that the sql returned is valid by using our sqlparser
2) check the query for keywords like `pre-reveal` instead of exact string matching

datadanne (commented):
Nice! I don't have a strong preference between this new query and the old one, I think they are both pretty easy to read.

If we think of Rigs as our goto showcase implementation for NFT projects, I wonder what the best query to show is? I think that the new query you wrote would feel intuitive for someone with decent SQL experience, but the old query might be more natural for someone that doesn't know SQL?

Fwiw `json_insert` is a variadic function so it could be extended pretty easily as well with 
```js
json_insert(
  json_array(...),
  '$[#]',
  json_object('key', 1),
  '$[#]',
  json_object('another-key', 2)
)
```
example (with `$`s escaped)
```
$ sqlite3 "" "select json_insert(json_array(json_object('a', 1)), '\$[#]', json_object('a', 2), '\$[#]', json_object('a', 3), '\$[#]', json_object('a', 4));" | jq
[
  {
    "a": 1
  },
  {
    "a": 2
  },
  {
    "a": 3
  },
  {
    "a": 4
  }
]
```

ethereum/helpers/uris.ts
asutula: Oh great call to use the parser here!

sanderpick: Yep, we'd have to change the regex to preserve single spaces across line breaks, and then add the single spaces at the end of each line that needs it.

For the first test:

Screenshot 2023-03-01 at 8 37 56 AM

For the second test:

Screenshot 2023-03-01 at 8 41 03 AM

The second test will still fail though because it looks like the expected case wasn't updated in the tests.

That would be the quick fix. But I agree w/ @datadanne that this is hard to maintain... classic case of needing to maintain complex expected results.

asutula:
Thanks for the thoughtful review @datadanne. I agree with you that it could go either way, but what pushed me in this direction is that my work on adding Filecoin deals to the metadata requires adding more than one additional attribute, but all sourced from the same table of deal information. It wasn't clear to me that you can insert multiple items to the json array without adding them individually, separated by `'$[#]',`, which would result in multiple identical subqueries each targeting some unique id. That starts to get awkward and may not work at all if you need to add an unknown number of additional attributes. Do you know if there is a way to support this easily/cleanly with `json_insert`?

sanderpick:
This is slick @asutula. Imo it is now easier to read and is a nice demo of `union` support. I suspect everyone will have a different take on what style is more readable though.

datadanne:
> Do you know if there is a way to support this easily/cleanly with json_insert?

I don't have a lot of experience with sqlite & json but I don't think so. There is a `json_merge` function in some other db implementations that supports concatenating arrays that would do it, but it doesn't look like sqlite supports that

ethereum/helpers/uris.ts
datadanne: > Oh great call to use the parser here!

Yeah it makes it much simpler and nicer :) Do you want to integrate that in this PR or should we file an issue and do that work later?

sanderpick: oops, type above... the regex is just `/( {2,}|\n)/g` (one space: match two or more, single newline: match)

asutula: Ok I went ahead with @datadanne's suggestion and updated the URI template generation and tests to use the wasm parser to normalize. Feels good to be able to use whatever formatting of sql that makes reading the easiest!

ethereum/helpers/uris.ts
```diff
@@ -1,72 +1,89 @@
...
   lookupsTable: string,
   pilotsTable: string,
   displayAttributes: boolean
-): string[] {
+): Promise<string[]> {
   if (!displayAttributes) {
     const uri =
       tablelandHost +
-      "/query?extract=true&unwrap=true&s=" +
+      "/api/v1/query?format=objects&extract=true&unwrap=true&statement=" +
```
asutula, line 20: Updated to the new API url. cc @joewagner

ethereum/test/TablelandRigs.ts
```diff
@@ -532,41 +532,39 @@ describe("Rigs", function () {
...
+        table2,
+        table3,
+        false
+      );
+      const uri = uriTemplate.join("1");
+      const uriParts = uri.split("statement=");
+      expect(uriParts.length).to.equal(2);
+      const statement = decodeURIComponent(uriParts[1]);
+      const normalizedStatement = await normalize(statement);
+      expect(normalizedStatement).contains("pre-reveal");
```
asutula, line 547: Updated test logic that does what @datadanne suggested and allows us to not have to maintain those complex test expectations.

asutula:
Thanks for the input guys. After the discussion with @datadanne, I do think using `union` is the way to go since it easily supports adding 0-n metadata attributes easily and cleanly, while SQLite's json functions don't seem to make that possible. Like I mentioned before, I'll be using this to add Filecoin deal data attributes to the metadata in the near future. I also updated the the metadata queries to use the new validator API like @joewagner brought up. Also updated the tests to work the way @datadanne suggested. Ready for a proper/final review here. Thanks!

ethereum/helpers/uris.ts
asutula: I think `format=objects` is the default, but added it just to be explicit and future proof since this ends up in smart contract storage.

datadanne: Could we use the sdk somehow to build this uri instead so that we don't have to remember to make updates here if the api changes? Might not be worth it since it probably won't change very often

ethereum/test/TablelandRigs.ts
```diff
@@ -532,41 +532,39 @@ describe("Rigs", function () {
...
+        tablelandHost,
+        table1,
+        table2,
+        table3,
+        false
+      );
+      const uri = uriTemplate.join("1");
+      const uriParts = uri.split("statement=");
+      expect(uriParts.length).to.equal(2);
+      const statement = decodeURIComponent(uriParts[1]);
```
datadanne, line 545:
```suggestion
      const uri = new URL(uriTemplate.join("1"));
      const statement = uri.searchParams.get("statement");
      // normalizing the statement checks that it is valid SQL, will throw
      // an error if it isn't
```
This is a super small nitpick but we could use the `URL` class here, might be a bit nicer? And a comment that mentions that calling `normalize` checks that the statement is valid SQL might make sense since we don't have an explicit assert for that?

datadanne (approved):
Nice!

ethereum/helpers/uris.ts
asutula: Took a quick look and I don't see anything useful from the sdk here.

datadanne (approved):
LGTM!

ethereum/helpers/uris.ts
joewagner: Sorry, I'm a little slow to respond here.  The way you did this works, so no need to change now, but you can potentially leverage the sdk's helpers if you know the chain being targeted.
example:
```
import { helpers } from "@tableland/sdk";

const tablelandHost = helpers.getChainInfo("homestead").baseUrl;
```

Issue #527, opened 2023-01-05
Rigs Contract: Implement on-chain metadata updates


dtbuchholz:
OpenSea now supports contract-triggered metadata updates: https://twitter.com/ralxzryan/status/1611058555247001601?s=46&t=8DHAlZH_0Qm51OxkZSfXTw

"EIP-4906 is a metadata update extension to ERC-721, with two events `MetadataUpdate` and `BulkMetadataUpdate`, that signal when a token should have its metadata refreshed": https://eips.ethereum.org/EIPS/eip-4906

sanderpick:
i added this one to Garage V2 project.

PR #688, opened 2023-03-09
feat: add EIP4906 metadata refresh

dtbuchholz:
- Emits a `MetadataUpdate` event with a `trainRig`, `parkRig`, `pilotRig`, or `parkRigAsOwner` action. This occurs on the `TablelandRigs` contract and not the `TablelandRigPilots`.
- Adds an additional check in `supportsInterface` with the IERC4906 inteface (0x49064906).
- Adds an additional contract in the `interface` directory for the `IERC4906.sol` metadata update.
- Includes a test for metadata update events.
- See the EIP for more details: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-4906.md
- Closes https://github.com/tablelandnetwork/rigs/issues/527

### DIFF
--- a/ethereum/contracts/interfaces/IERC4906.sol
+++ b/ethereum/contracts/interfaces/IERC4906.sol
@@ -1,9 +1,6 @@
 // SPDX-License-Identifier: Unlicense
 pragma solidity >=0.8.10 <0.9.0;
 
-import "@openzeppelin/contracts/interfaces/IERC165.sol";
-import "@openzeppelin/contracts/interfaces/IERC721.sol";
-
 /**
  * @dev Interface for IERC4906 (EIP-721 Metadata Update Extension).
  */
--- a/ethereum/contracts/interfaces/IERC4906.sol
+++ b/ethereum/contracts/interfaces/IERC4906.sol
@@ -1,0 +1,24 @@
+// SPDX-License-Identifier: Unlicense
+pragma solidity >=0.8.10 <0.9.0;
+
+import "@openzeppelin/contracts/interfaces/IERC165.sol";
+import "@openzeppelin/contracts/interfaces/IERC721.sol";
+
+/**
+ * @dev Interface for IERC4906 (EIP-721 Metadata Update Extension).
+ */
+interface IERC4906 {
+    /**
+     * @dev This event emits when the metadata of a token is changed.
+     * So that the third-party platforms such as NFT market could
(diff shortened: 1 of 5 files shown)


ethereum/contracts/interfaces/IERC4906.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: Unlicense
```
dtbuchholz, line 1: OZ has the contract but it's not available via npm, yet. https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/interfaces/IERC4906.sol.

Since the Rigs contract already inherits from those that implement IERC165 and IERC721, I didn't have IERC4906 inherit them here (double checking this is fine as it differs from the contract linked above).

ethereum/test/TablelandRigs.ts
```diff
@@ -1944,5 +1946,60 @@ describe("Rigs", function () {
           );
       });
     });
+
+    describe("metadata refresh", function () {
```
sanderpick, line 1950: Nit: add a test for parking as rig owner

ethereum/contracts/interfaces/IERC4906.sol
```diff
@@ -0,0 +1,24 @@
...
+pragma solidity >=0.8.10 <0.9.0;
+
+import "@openzeppelin/contracts/interfaces/IERC165.sol";
+import "@openzeppelin/contracts/interfaces/IERC721.sol";
```
sanderpick, line 5: Are these used?

sanderpick (commented):
looks great! I left a couple comments

ethereum/contracts/interfaces/IERC4906.sol
dtbuchholz: ah, no—they were part of the original OZ interface but should have removed

sanderpick (approved):
Sweet. LGTM!

dtbuchholz:
@sanderpick Since I'm the Mumbai contract owner, should I redeploy the Rigs contract with these changes before merging? It's been a while, so I wanna double check that all I need to do is run the `upgradeRigs.ts` script, right?

sanderpick:
Nice yeah, good idea. Correct, if pilots is unchanged, just need to upgrade Rigs.

sanderpick (approved):

PR #714, opened 2023-03-24
Contracts: Add parking admin

datadanne:
Adds a role/address that is allowed to force park rigs, and the address can be set by the contract owner. 

I wonder if I should update the naming to just `admin` or maybe `rigsAdmin` instead of `parkingAdmin` so that we can allow the admin to perform other actions that aren't tied to parking in the future without having to change the name? Or maybe having to use separate wallets for each type of task is a good design so that we can allow some users to perform a single admin action, kind of like rbac

Closes RIG-13

### DIFF
--- a/ethereum/helpers/uris.ts
+++ b/ethereum/helpers/uris.ts
@@ -22,7 +22,7 @@ export async function getURITemplate(
         await normalize(
           `select 
             json_object(
-              'name','Rig #'||rig_id,
+              'name',case when exists(select * from ${pilotsTable} where rig_id = ${attributesTable}.rig_id and end_time is null) then 'Rig #'||rig_id||' ✈️' else 'Rig #'||rig_id end,
               'external_url','https://garage.tableland.xyz/rigs/'||rig_id,
               'image','ipfs://'||renders_cid||'/'||rig_id||'/'||image_full_name,
               'image_alpha','ipfs://'||renders_cid||'/'||rig_id||'/'||image_full_alpha_name,
@@ -54,7 +54,7 @@ export async function getURITemplate(
         await normalize(
           `select
             json_object(
-              'name','Rig #'||rig_id,
+              'name',case when exists(select * from ${pilotsTable} where rig_id = result.rig_id and end_time is null) then 'Rig #'||rig_id||' ✈️' else 'Rig #'||rig_id end,
               'external_url','https://garage.tableland.xyz/rigs/'||rig_id,
               'image','ipfs://'||renders_cid||'/'||rig_id||'/'||image_full_name,
               'image_alpha','ipfs://'||renders_cid||'/'||rig_id||'/'||image_full_alpha_name,
@@ -80,7 +80,7 @@ export async function getURITemplate(
               from
                 ${attributesTable} a
                 left join (select * from ${pilotsTable} where end_time is null) s on a.rig_id = s.rig_id
-            )
+            ) as result
             join ${lookupsTable}
           where rig_id=ID
           group by rig_id;`


ethereum/contracts/TablelandRigs.sol
```diff
@@ -507,6 +545,25 @@ contract TablelandRigs is
         }
     }
 
+    /**
+     * @dev See {ITablelandRigs-parkRigAsParkingAdmin}.
+     */
+    function parkRigAsParkingAdmin(uint256[] calldata tokenIds) external onlyParkingAdmin {
```
sanderpick, line 551: To save contract size, looks like you could make a private `_parkAsOwnerOrAdmin` method and call it from `parkRigAsParkingAdmin` and `parkRigAsOwner`, which are gated by respective modifiers.

sanderpick: Alternatively, you could have a modifier that is `onlyOwnerOrAdmin`, and a single external method `parkRigAsOwnerOrAdmin` (I don't think we need to worry about breaking changes for owner only methods).

sanderpick (commented):
Sweet! Left a couple comments.

ethereum/contracts/ITablelandRigs.sol
```diff
@@ -158,6 +158,22 @@ interface ITablelandRigs {
      */
     function setRoyaltyReceiver(address receiver) external;
 
+    /**
+     * @dev Returns the parking admin.
+     */
+    function parkingAdmin() external view returns (address);
```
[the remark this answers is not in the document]
datadanne: Changed to admin

ethereum/contracts/TablelandRigs.sol
datadanne: I added a private function that is called from both the Owner and Admin park functions

datadanne:
> Sweet! Left a couple comments. 

Thanks! Made some changes based on your comments 👍

ethereum/test/TablelandRigs.ts
```diff
@@ -1781,10 +1801,6 @@ describe("Rigs", function () {
       it("Should block contract non-owner", async function () {
         const _rigs = rigs.connect(accounts[2]);
 
-        await expect(
-          _rigs.initPilots(ethers.constants.AddressZero)
-        ).to.be.rejectedWith("Ownable: caller is not the owner");
```
sanderpick, line 1786: Nice, guessing this is a no-op

ethereum/contracts/TablelandRigs.sol
```diff
@@ -352,6 +369,30 @@ contract TablelandRigs is
...
+    modifier onlyAdmin() {
+        _checkAdmin();
+        _;
+    }
+
+    /**
+     * @dev Throws if the sender is not the admin or if admin has not
+     * been initialized.
+     */
+    function _checkAdmin() internal view virtual {
```
sanderpick, line 388: Looks like this could be `private view`. It doesn't need to be overridden by some inheriting contract like in `Ownable`.

ethereum/contracts/TablelandRigs.sol
```diff
@@ -488,10 +529,7 @@ contract TablelandRigs is
         }
     }
 
-    /**
-     * @dev See {ITablelandRigs-parkRigAsOwner}.
-     */
-    function parkRigAsOwner(uint256[] calldata tokenIds) external onlyOwner {
+    function _forceParkRigs(uint256[] calldata tokenIds) internal {
```
sanderpick, line 532: Looks like this can be `private`.

sanderpick (commented):
Looks good! I left a couple nits on function visibility.

ethereum/test/TablelandRigs.ts
datadanne: Yeah, I removed it because it is irrelevant to this test case, probably because the test case was copy/pasted. It is covered in another test case!

sanderpick (approved):
👍

PR #720, opened 2023-03-28
Fix training state mistmatch

dtbuchholz:
This PR fixes an issue where contract and table state diverged due to logic in `parkRig`. It also adds an overloaded `pilotInfo` that can be called with multiple Rig IDs.

The original issue boils down to the following in `parkRig` where an initial `trainingIncomplete` check would `_setPilotData`, separate from the ensuing `UPDATE` statement. 
```solidity
// Pilot training is incomplete; reset the training pilot such that the Rig must train again
if (trainingIncomplete) _setPilotData(uint16(tokenId), 0, 0);
// If the pilot is untrained or being force parked, there are zero flight time rewards
if (trainingIncomplete || force) {
    // Update the row in pilot sessions table with its `end_time` equal to the `start_time`
    setters = string.concat(
        "end_time=",
        StringsUpgradeable.toString(startTime)
    );
} else {
  // ...
}
```
The contract and table state are now synced:
```solidity
if (trainingIncomplete || force) {
    _setPilotData(uint16(tokenId), 0, 0);
    setters = string.concat(
        "end_time=",
        StringsUpgradeable.toString(startTime)
    );
}
```
The one callout is that a force parked Rig will revert to a training status and have to go through training again, even it was previously able to pilot the Rig. Lore-wise with these changes, a Rig is sentient in that it can be trained to handle a pilot, but if the pilot acts maliciously (i.e., tries to sell the Rig while in flight), the Rig loses training knowledge and must re-train.

### DIFF
--- a/ethereum/contracts/TablelandRigPilots.sol
+++ b/ethereum/contracts/TablelandRigPilots.sol
@@ -282,6 +282,50 @@ contract TablelandRigPilots is
     /**
      * @dev See {ITablelandRigPilots-pilotRig}.
      */
+    function pilotRig(address sender, uint256 tokenId) external onlyParent {
+        // Validate the Rig can be piloted with a trainer pilot. Note that
+        // `_canPilot` allows for in-flight piloting once trained *and* in
+        // `TRAINING` status, but this only allows for piloting a with trainer
+        // if `PARKED`.
+        if (_pilotStatus(tokenId) != GarageStatus.PARKED)
+            revert InvalidPilotStatus();
+
+        // Assign a trainer pilot to the Rig in `_pilots`. The pilot contract is
+        // `0` and pilot ID `2`—this is needed to differentiate from a
+        // `TRAINING` Rig's pilot, which has a pilot ID of `1`.
+        _setPilotData(uint16(tokenId), 0, 2);
+
+        // Set the start time for the new pilot session
+        _setStartTime(uint16(tokenId), uint64(block.number));
+
+        // Insert the trainer pilot into the Tableland pilot sessions table
+        TablelandDeployments.get().runSQL(
+            address(this),
+            _pilotSessionsTableId,
+            SQLHelpers.toInsert(
+                _PILOT_SESSIONS_PREFIX,
+                _pilotSessionsTableId,
+                "rig_id,owner,pilot_contract,pilot_id,start_time",
+                string.concat(
+                    StringsUpgradeable.toString(uint16(tokenId)),
+                    ",",
+                    SQLHelpers.quote(StringsUpgradeable.toHexString(sender)),
+                    ",",
+                    SQLHelpers.quote(Strings.toHexString(address(0))),
+                    ",",
+                    StringsUpgradeable.toString(uint32(2)),
+                    ",",
+                    StringsUpgradeable.toString(uint64(block.number))
+                )
+            )
+        );
+
+        emit Piloted(tokenId, address(0), 2);
+    }
+
+    /**
+     * @dev See {ITablelandRigPilots-pilotRig}.
+     */
     function pilotRig(
         address sender,
         uint256 tokenId,
@@ -305,7 +349,8 @@ contract TablelandRigPilots is
         // Validate the Rig can be piloted
         if (!_canPilot(tokenId)) revert InvalidPilotStatus();
 
-        // Initialize the packed "pilot data" (pilot ID `uint32` with a pilot contract `uint160`, shifted 32 bits)
+        // Initialize the packed "pilot data" (pilot ID `uint32` with a pilot
+        // contract `uint160`, shifted 32 bits)
         uint192 pilotData = (uint192(pilotId) |
             (uint192(uint160(pilotAddr)) << 32));
 
@@ -320,8 +365,9 @@ contract TablelandRigPilots is
             _pilotStatus(_pilotIndex[pilotData]) != GarageStatus.PARKED
         ) parkRig(_pilotIndex[pilotData], false);
 
-        // If the Rig is training, end its training session (no parking required) to then open a new pilot session
-        // Pilot has completed training at this point (training validation is checked above)
+        // If the Rig is training, end its training session (no parking
+        // required) to then open a new pilot session. Pilot has completed
+        // training at this point (training validation is checked above)
         if (_pilotStatus(tokenId) == GarageStatus.TRAINING) {
             // Update the pilot's existing training session with its `end_time`
             string memory setters = string.concat(
--- a/ethereum/contracts/TablelandRigPilots.sol
+++ b/ethereum/contracts/TablelandRigPilots.sol
@@ -298,22 +298,18 @@ contract TablelandRigPilots is
         // Set the start time for the new pilot session
         _setStartTime(uint16(tokenId), uint64(block.number));
 
-        // Insert the trainer pilot into the Tableland pilot sessions table
+        // Insert the trainer pilot session into the Tableland pilot sessions table
         TablelandDeployments.get().runSQL(
             address(this),
             _pilotSessionsTableId,
             SQLHelpers.toInsert(
                 _PILOT_SESSIONS_PREFIX,
                 _pilotSessionsTableId,
-                "rig_id,owner,pilot_contract,pilot_id,start_time",
+                "rig_id,owner,start_time",
                 string.concat(
                     StringsUpgradeable.toString(uint16(tokenId)),
                     ",",
                     SQLHelpers.quote(StringsUpgradeable.toHexString(sender)),
-                    ",",
-                    SQLHelpers.quote(Strings.toHexString(address(0))),
-                    ",",
-                    StringsUpgradeable.toString(uint32(2)),
                     ",",
                     StringsUpgradeable.toString(uint64(block.number))
                 )
--- a/ethereum/contracts/TablelandRigPilots.sol
+++ b/ethereum/contracts/TablelandRigPilots.sol
@@ -395,10 +395,13 @@ contract TablelandRigPilots is
         uint64 startTime = pilotStartTime(tokenId);
         bool trainingIncomplete = status == GarageStatus.TRAINING &&
             !(block.number >= startTime + _PILOT_TRAINING_DURATION);
+        bool forceParkedWhileTraining = status == GarageStatus.TRAINING &&
+            force;
+        // Pilot training is incomplete; reset the training pilot such that the Rig must train again
+        if (trainingIncomplete || forceParkedWhileTraining)
+            _setPilotData(uint16(tokenId), 0, 0);
         // If the pilot is untrained or being force parked, there are zero flight time rewards
         if (trainingIncomplete || force) {
-            // Reset the training pilot such that the Rig must train again
-            _setPilotData(uint16(tokenId), 0, 0);
             // Update the row in pilot sessions table with its `end_time` equal to the `start_time`
             setters = string.concat(
                 "end_time=",
--- a/ethereum/contracts/TablelandRigPilots.sol
+++ b/ethereum/contracts/TablelandRigPilots.sol
@@ -395,10 +395,10 @@ contract TablelandRigPilots is
         uint64 startTime = pilotStartTime(tokenId);
         bool trainingIncomplete = status == GarageStatus.TRAINING &&
             !(block.number >= startTime + _PILOT_TRAINING_DURATION);
-        // Pilot training is incomplete; reset the training pilot such that the Rig must train again
-        if (trainingIncomplete) _setPilotData(uint16(tokenId), 0, 0);
         // If the pilot is untrained or being force parked, there are zero flight time rewards
         if (trainingIncomplete || force) {
+            // Reset the training pilot such that the Rig must train again
+            _setPilotData(uint16(tokenId), 0, 0);
             // Update the row in pilot sessions table with its `end_time` equal to the `start_time`
             setters = string.concat(
                 "end_time=",
--- a/ethereum/contracts/TablelandRigs.sol
+++ b/ethereum/contracts/TablelandRigs.sol
@@ -412,15 +412,32 @@ contract TablelandRigs is
     }
 
     /**
-     * @dev See {ITablelandRigs-pilotInfo}.
+     * @dev See {ITablelandRigs-pilotInfo(uint256)}.
      */
     function pilotInfo(
         uint256 tokenId
-    ) external view returns (ITablelandRigPilots.PilotInfo memory) {
+    ) public view returns (ITablelandRigPilots.PilotInfo memory) {
         // Check the Rig `tokenId` exists
         if (!_exists(tokenId)) revert OwnerQueryForNonexistentToken();
 
         return _pilots.pilotInfo(tokenId);
+    }
+
+    /**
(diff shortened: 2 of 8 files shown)


dtbuchholz:
@datadanne re: the side effect—agreed. I realized I misunderstood the root cause of this issue and will revert that change because the original thinking is the right approach.

Maybe instead of altering `trainRig`, it stays the same but we either:
1. Add an overloaded `pilotRig` that doesn't take a custom `pilotAddr` nor `pilotId`, thus, assuming a trainer pilot.
2. Change the existing `pilotRig` to allow `0` values for the `pilotAddr` and `pilotId`, indicating a trainer pilot.

Thus, the `trainRig` / training logic and everything else stays the same. For reference, the fn takes the following:

```solidity
function pilotRig(
    address sender,
    uint256 tokenId,
    address pilotAddr,
    uint256 pilotId
) external;
```

datadanne:
Overloading or allowing 0 values for the pilot method both sgtm!

I think the change you made to parkRig makes sense as well for the case where a rig that is in-flight on its first session (> 172,800 blocks) gets force parked, in that case I don't think we want the rig to be marked as trained in the contract (and that is an inconsistency between the state in the contract and the state in tableland since we have 0FT recorded in the table but the rig is still marked as trained). But I don't think we should mark rigs that are on their second or later session and get force parked as untrained which I think the current version of the code in this PR does? That would introduce another inconsistency between the state in the contract and tableland, the rig could have lots of FT recorded in tableland but would be marked as untrained in the contract

Does that make sense @dtbuchholz? Cc @sanderpick

sanderpick:
@datadanne reading here: 

> if a rig that has been training for more than 172,800 blocks gets force parked we mark it as trained in the contract but set the pilot session length to 0. if this has happened the garage thinks the rig is untrained and shows a train button, but that transaction will get reverted

In the case of "a rig that has been training for more than 172,800 blocks gets force parked", the `GarageStatus` will be `PARKED` right? Could you use `PilotInfo.status` to drive the UI instead of pilot session length? ie, 

Show "train" button if:
- status = `UNTRAINED`

Show pilot button if:
- status = `PARKED`

sanderpick:
Ah, nvm, catching up. I see from another convo that you don't want the garage app to have to call a method on the contract. Makes sense.

sanderpick:
So tl;dr, we want to reset a rig to `UNTRAINED` and award zero FT for the active session if they are being forced parked, regardless of how long they have been training.

@dtbuchholz, I think the `parkRig` logic needs to be something like:

```solidity
// Session update type is dependent on training completion status
// Use `setters` for setting SQL update values, and `startTime` is used for checking training completion status
string memory setters;
uint64 startTime = pilotStartTime(tokenId);
bool trainingIncomplete = status == GarageStatus.TRAINING &&
    !(block.number >= startTime + _PILOT_TRAINING_DURATION);
bool punishTrainer = status == GarageStatus.TRAINING && force;
// Pilot training is incomplete; reset the training pilot such that the Rig must train again
if (trainingIncomplete || punishTrainer)
    _setPilotData(uint16(tokenId), 0, 0);
// If the pilot is untrained or being force parked, there are zero flight time rewards
if (trainingIncomplete || force) {
    // Update the row in pilot sessions table with its `end_time` equal to the `start_time`
    setters = string.concat(
        "end_time=",
        StringsUpgradeable.toString(startTime)
    );
} else {
    // Training is complete, and the Rig is being parked without "malicious" force park origins
    // Update the row in the pilot sessions table with its `end_time` to accrue flight time rewards
    setters = string.concat(
        "end_time=",
        StringsUpgradeable.toString(uint64(block.number))
    );
}
```

because we don't want to reset _all_ rigs that are forced parked to `UNTRAINED`, just ones that are being `TRAINING` when force parked. 

I like your option 2 to fix the broken state of some rigs from the garage perspective:

> 2. Change the existing pilotRig to allow 0 values for the pilotAddr and pilotId, indicating a trainer pilot.

sanderpick:
(no need to use `punishTrainer` variable, just illustrating the idea)

datadanne:
> So tl;dr, we want to reset a rig to UNTRAINED and award zero FT for the active session if they are being forced parked, regardless of how long they have been training.

I think this makes the most sense!

dtbuchholz:
Ok cool, makes sense, thanks. Will implement this now.

ethereum/contracts/TablelandRigs.sol
```diff
@@ -425,7 +442,17 @@ contract TablelandRigs is
         if (ownerOf(tokenId) != _msgSenderERC721A())
             revert ITablelandRigPilots.Unauthorized();
 
-        _pilots.pilotRig(_msgSenderERC721A(), tokenId, pilotAddr, pilotId);
+        // If the supplied pilot address is `0x0`, then assume a trainer pilot
+        // (note: `pilotId` has no impact here). Otherwise, proceed with a
+        // custom pilot. The overloaded methods direct changes accordingly.
+        pilotAddr == address(0)
```
dtbuchholz, line 448: This is the only logic update in the Rigs contract, so there aren't any fn signature changes. You can now pass `0x0` as contract value, which will assume you're trying to pilot your Rig with a "stock" trainer pilot. 

In this situation, only the passed `pilotAddr` matters; `pilotId` has no impact within the Rigs contract itself but is hardcoded in the `TablelandRigPilots` contract to a value of `2`.

ethereum/contracts/TablelandRigPilots.sol
```diff
@@ -279,6 +279,50 @@ contract TablelandRigPilots is
         emit Training(tokenId);
     }
 
+    /**
+     * @dev See {ITablelandRigPilots-pilotRig}.
+     */
+    function pilotRig(address sender, uint256 tokenId) external onlyParent {
```
dtbuchholz, line 285: An overloaded `pilotRig` doesn't take any pilot contract nor ID values and will assume a `0x0` address with pilot ID hardcoded as `2`.

Only a parked Rig can have a trainer pilot set with this method. It will insert these values into the sessions table, which also helps ensure training-related sessions differ from piloting with a stock / trainer pilot.

ethereum/test/TablelandRigs.ts
```diff
@@ -1831,6 +2067,109 @@ describe("Rigs", function () {
...
+        await network.provider.send("hardhat_mine", [
+          ethers.utils.hexValue(172800),
+        ]);
+        // Park the Rig as the contract owner
+        await rigs.parkRigAsOwner([BigNumber.from(tokenId)]);
+        // Check pilot is back to untrained
+        const pilotInfo = await rigs["pilotInfo(uint256)"](
+          BigNumber.from(tokenId)
+        );
+        // Check that the pilot is now `0` since training was incomplete
```
sanderpick, line 2094: nit on the comment: Check that the pilot is now `0` since it was force parked while training

sanderpick (approved):
LGTM! Would be good to wait for @datadanne to have a look as well.

sanderpick:
oop, didn't mean to approve, just comment. blocked anyway by tests.

ethereum/test/TablelandRigs.ts
dtbuchholz: oops, didn't see this, fixed!

ethereum/contracts/TablelandRigPilots.sol
```diff
@@ -279,6 +279,50 @@ contract TablelandRigPilots is
...
+                    StringsUpgradeable.toString(uint16(tokenId)),
+                    ",",
+                    SQLHelpers.quote(StringsUpgradeable.toHexString(sender)),
+                    ",",
+                    SQLHelpers.quote(Strings.toHexString(address(0))),
+                    ",",
+                    StringsUpgradeable.toString(uint32(2)),
+                    ",",
+                    StringsUpgradeable.toString(uint64(block.number))
+                )
```
datadanne, line 319: `trainRig` doesn't write a `pilot_contract` or `pilot_id` value to the pilot sessions table so the data will look pretty different between a training session and a non-training session without an NFT pilot.

If we want to think of training and flying without an NFT pilot as using two separate pilots I think this makes sense, I think we should probably set the contract value to null and not 0x0 to be consistent with the training session but that is a minor detail. And we would have to update the garage to know that it shouldn't try to fetch NFT metadata for the collection `{ contract: 0x0, id: 2 }` and that we should hardcode that as "Default Pilot" instead of "Trainer" or something like that

But if we don't want to think of it as two separate pilots I think this is a suboptimal solution, working with the data would be a bit annoying since you can no longer do a simple
```sql
SELECT pilot_contract, pilot_id, count(*) FROM pilot_sessions WHERE rig_id = ?id GROUP BY pilot_contract, pilot_id
```
to get the number of flight sessions per pilot for instance. In that case I think we should either add a column to the pilot_sessions table to indicate what type of session it is (training/not training) and use the same contract and id (which would be annoying since we don't support migrations), or add another table called something like pilot_session_type where we store `(pilot_session_id, type)` tuples

Does that make sense?

datadanne (commented):
Nice! Have a question re using `contract: 0x0, id: 2` as the session data

ethereum/contracts/TablelandRigPilots.sol
sanderpick: @datadanne would the garage require changes if we simply didn't write `pilot_contract` and `pilot_id` to the pilots table when a rig that was `PARKED` goes flying with the trainer? ie, in `function pilotRig(address sender, uint256 tokenId) external onlyParent` do:

```solidity
TablelandDeployments.get().runSQL(
    address(this),
    _pilotSessionsTableId,
    SQLHelpers.toInsert(
        _PILOT_SESSIONS_PREFIX,
        _pilotSessionsTableId,
        "rig_id,owner,start_time",
        string.concat(
            StringsUpgradeable.toString(uint16(tokenId)),
            ",",
            SQLHelpers.quote(StringsUpgradeable.toHexString(sender)),
            ",",
            StringsUpgradeable.toString(uint64(block.number))
        )
    )
);
```

Then, from the garage perspective, when you pilot with the trainer by clicking "pilot", you can treat it like it like normal training. One hiccup I can think of is how to know if the rig can be piloted from training. This is how the contract determines if this is possible:

```
// Cannot switch real pilots mid-flight, but can go from trainer -> real pilot
// There are two ways to pilot:
//   1. In a `PARKED` status
//   2. In a `TRAINING` status for long enough (30 days; 172800 blocks)
```

So, you'd have to determine if the rig has previously completed a training session, which I think you already do by querying for at least one completed session. You'd have to grey out the pilot button when the trainer is in use in this case. That might be confusing, but not too bad.

dtbuchholz: (I'm fine with `NULL` pilot contract/ID values for any type of trainer pilot session—just lmk what to adjust here)

I didn't want to complicate things for this fix but had this broader feeling: a trainer pilot _is_ different than this "stock" default pilot. We _could_ make the stock pilot a Rig-bound token...nothing fancy with the token itself (e.g., same table setup as Rigs but basically all empty with a stock pilot token ID, one default attribute, and same rendering as the trainer but a different color). That bound stock pilot could be upgraded to something else in the future. Plus, that solves for the `0x0` and default `2` design since here, there would be a real pilot contract and unique pilot ID.

It's out of scope and wanted to throw the idea out there...but short term, not feasible.

sanderpick: sg. let's see what @datadanne says about my suggestion above.

I like the stock pilot idea... worth exploring the idea in a notion doc at some point?

dtbuchholz: Threw some ideas down here—not an official proposal but wanted to put it down in writing https://www.notion.so/textile/Rigs-stock-default-pilot-b88de504639448bfb98079a25805a6a8?pvs=4

datadanne: > @datadanne would the garage require changes if we simply didn't write pilot_contract and pilot_id to the pilots table when a rig that was PARKED goes flying with the trainer? ie, in function pilotRig(address sender, uint256 tokenId) external onlyParent do:

Nope, all the pilot-related code in the garage can stay the same if we leave the `pilot_contract` and `pilot_id` values empty when an already trained rig is piloted with the trainer pilot again. 

We still need to update the garage to allow a user to take a rig in-flight that is trained using the pilot, right now we only support using an ERC721 as the pilot if the rig is trained.

> ie, with @dtbuchholz 's change, the contract would still know about this distinction: TRAINING vs. PILOTED, which is used to determine if the rigs gets reset to UNTRAINED during a force park. But, i'm not sure the garage needs to know about that distinction at all.

Yeah the garage doesn't need to know this at all as long as rigs are reset to UNTRAINED when they get force parked in the TRAINING state. 

We will still need to look into how many rigs are in this inconsistent state right now though, if there are many that got force parked during training we still have to handle that in the garage even though the contract won't allow new rigs to get into that state

sanderpick: sg. @dtbuchholz, sounds like this PR will be good if you make the change here: https://github.com/tablelandnetwork/rigs/pull/720#discussion_r1154679035

dtbuchholz: @sanderpick I figure I should upgrade the mumbai contracts in this PR, too. With the changes to `upgradeRigs.ts` in https://github.com/tablelandnetwork/rigs/commit/024d569e8e8df1b63a6296d7ed65cef5e91beaee, if/how does that impact how I should upgrade rigs on mumbai? Should I ignore that `TMP` section when running the script?

sanderpick:
You shouldn’t need to ignore it. It’s just initializing the impl

dtbuchholz:
jfyi, that line ended up throwing this error but looks like it all deployed fine:
```
npx hardhat run scripts/upgradeRigs.ts --network polygon-mumbai
Compiled 72 Solidity files successfully

Upgrading on 'polygon-mumbai'...
Using address '0xfdD15E0Af08AF6b767bc85384698eda79f5AcAb4'
Current implementation address: 0x5Cc6f308381994F35c580eD049CAfCB684436736
New implementation address: 0x3989fBf8311ff6B9895ed397981DEF0055232b04
HeadersTimeoutError: Headers Timeout Error
    at Timeout.onParserTimeout [as _onTimeout] (/Users/dtb/tbl/rigs/ethereum/node_modules/undici/lib/client.js:894:28)
    at listOnTimeout (node:internal/timers:566:11)
    at processTimers (node:internal/timers:507:7) {
  code: 'UND_ERR_HEADERS_TIMEOUT'
}
```

sanderpick:
Hmm, that looks like a network timeout, maybe related to VPN? https://github.com/NomicFoundation/hardhat/issues/2672#issuecomment-1302021778

ethereum/deployments.ts
```diff
@@ -52,7 +52,7 @@ export const deployments: RigsDeployments = {
     royaltyContractAddress: "0xb61974afD4348DA16e45BC48d53883A281bc4A6e",
     pilotsAddress: "0x864655946432f52AE16BDE9af054380b3dE06789",
     tablelandChain: "maticmum",
-    tablelandHost: "https://testnet.tableland.network",
+    tablelandHost: "https://testnets.tableland.network",
```
sanderpick, line 55: good catch

sanderpick (approved):
Looks good! I can deploy this today after it's merged.

PR #729, opened 2023-04-02
Add delegate.cash support

datadanne:
Updates the contract to allow a delegated wallet to train/pilot/park rigs using delegate.cash.
Adds a "Act as" button to the garage top header that is only shown if the currently connected wallet is registered as a delegate for a vault for the Rigs contract. A click on the button lets the user change wallet that they want to make actions for:

image
image

Closes RIG-17

### DIFF
--- a/ethereum/contracts/DelegateCashMock.sol
+++ b/ethereum/contracts/DelegateCashMock.sol
@@ -1,23 +1,9 @@
 // SPDX-License-Identifier: Unlicense
 pragma solidity >=0.8.10 <0.9.0;
 
-interface MockDelegateCashInterface {
-    function checkDelegateForToken(
-        address delegate,
-        address vault,
-        address contract_,
-        uint256 tokenId
-    ) external view returns (bool);
+import "./interfaces/IDelegationRegistry.sol";
 
-    function registerDelegate(
-        address delegate,
-        address vault,
-        address contract_,
-        uint256 tokenId
-    ) external;
-}
-
-contract DelegateCashMock is MockDelegateCashInterface {
+contract DelegateCashMock is IDelegationRegistry {
     mapping(address => mapping(address => mapping(uint256 => mapping(address => bool)))) delegations;
 
     function checkDelegateForToken(
--- a/ethereum/contracts/DelegateCashMock.sol
+++ b/ethereum/contracts/DelegateCashMock.sol
@@ -1,0 +1,40 @@
+// SPDX-License-Identifier: Unlicense
+pragma solidity >=0.8.10 <0.9.0;
+
+interface MockDelegateCashInterface {
+    function checkDelegateForToken(
+        address delegate,
+        address vault,
+        address contract_,
+        uint256 tokenId
+    ) external view returns (bool);
+
+    function registerDelegate(
+        address delegate,
+        address vault,
+        address contract_,
+        uint256 tokenId
+    ) external;
+}
+
+contract DelegateCashMock is MockDelegateCashInterface {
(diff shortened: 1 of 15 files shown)


ethereum/contracts/TablelandRigs.sol
```diff
@@ -446,11 +464,25 @@ contract TablelandRigs is
...
+                tokenOwner,
+                address(this),
+                tokenId
+            )
+        ) {
+            sender = tokenOwner;
+        }
 
-        _pilots.trainRig(_msgSenderERC721A(), tokenId);
+        if (tokenOwner != sender) revert ITablelandRigPilots.Unauthorized();
```
sanderpick, line 483: this could be a function/modifier to save on contract size since it's reused three times

ethereum/contracts/DelegateCashMock.sol
```diff
@@ -0,0 +1,40 @@
...
+
+    function registerDelegate(
+        address delegate,
+        address vault,
+        address contract_,
+        uint256 tokenId
+    ) external;
+}
+
+contract DelegateCashMock is MockDelegateCashInterface {
```
sanderpick, line 20: so is there one "real" public delegatecash contract out in the wild that we'll use?

datadanne: Yeah. delegate.cash is deployed to the same address on most networks (https://docs.delegate.cash/delegatecash/technical-documentation/delegation-registry/contract-addresses), but we can't add new constants to the contract since it is upgradeable so I think we need to use a setter instead?

sanderpick: we could add new consts... I think it's just a matter of if we want to. may as well to avoid a new method? note: consts can change with upgrades since they don't use contract storage.

datadanne: yeah I agree. I was going to do that first but for some reason I thought that we couldn't add a constant to an upgradeable contract. fixed!

ethereum/contracts/interfaces/IDelegationRegistry.sol
datadanne, line 1: The full implementation is available here https://docs.delegate.cash/delegatecash/technical-documentation/delegation-registry/idelegationregistry.sol, but since we only use one method it felt simpler for testing purposes to use a reduced version. Do you have a preference or do you know if there are any best practices @sanderpick?

sanderpick: What you did (the minimal interface) looks! I've seen that pattern elsewhere... not sure about solidity best practice there but in general software eng, declaring the subset of an interface that you're using close to the impl is a healthy practice.

sanderpick:
@datadanne it would be good to get this PR resolved this week!

datadanne:
> @datadanne it would be good to get this PR resolved this week!

Agreed! I think I addressed all your comments last week, and I just rebased it so it should be ready for review again.

sanderpick:
I realized I should have merged this on Friday since you were out. We're gonna try merging normal work into a staging branch to avoid dangling PRs. I'll create one on this repo. We can merge to there and ship it next week. 

Bug fixe PRs can just be merged to main.

datadanne:
> I realized I should have merged this on Friday since you were out. We're gonna try merging normal work into a staging branch to avoid dangling PRs. I'll create one on this repo. We can merge to there and ship it next week.
> 
> Bug fixe PRs can just be merged to main.

Sounds good! And it kind of worked out well that none of us merged it last week since it would have deployed the UI change right away, but the UI won't work until we've deployed the contract update 👍

PR #735, opened 2023-04-04
Garage: Add RigDetails train/pilot fix for rigs in inconsistent state

datadanne:
Updates the rig details page to make an on-chain call to `pilotInfo` and to take the contract state into account when deciding which train/pilot button to show. This PR doesn't change the dashboard so it will still show the incorrect action (train instead of pilot) if you select a rig in an inconsistent state, and batch piloting one of those rigs won't work.

Since we know that we only have 20 rigs in this state and we can get 9 fixed by working with one user I think this should be an OK solution. What do you think @sanderpick?

### DIFF
--- a/garage/src/pages/RigDetails/index.tsx
+++ b/garage/src/pages/RigDetails/index.tsx
@@ -40,7 +40,7 @@ import { prettyNumber, truncateWalletAddress } from "../../utils/fmt";
 import { sleep } from "../../utils/async";
 import { chain, openseaBaseUrl, deployment } from "../../env";
 import { RigWithPilots, isValidAddress } from "../../types";
-import { abi } from "../../abis/ERC721";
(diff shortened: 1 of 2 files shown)


sanderpick (approved):
lgtm! i guess we can remove this logic if all those stuck rigs get unstuck.

sanderpick:
- Dan's fix has been deployed
- You might want the latest rigs package: https://github.com/tablelandnetwork/rigs/pull/740

datadanne:
> * Dan's fix has been deployed
> * You might want the latest rigs package: [Publish and bump rigs in animation_url and garage #740](https://github.com/tablelandnetwork/rigs/pull/740)

Nice! I can rebase this PR as soon as you merge #740

datadanne:
Rebased! Should be ready to go

PR #781, opened 2023-04-17
Update Tableland Go client

asutula:
This is in theory pretty straight forward, but uncovered this problem with tidying Go dependencies https://github.com/tablelandnetwork/go-tableland/issues/531. I'd like to see if we can get that resolved before merging this, so leaving this in draft for now.


asutula:
Ok go that dependency issue resolved. This is ready now.

sanderpick (approved):
Nice, looks good to me!

PR #811, opened 2023-04-27
Garage: fix wallet disconnect (update rainbowkit & wagmi to 0.12)

datadanne:
Metamask doesn't allow you to programmatically disconnect a wallet, the user needs to disconnect it themselves. But wagmi has a `shimDisconnect` option that keeps track of the disconnect state locally in the connector using local storage. RainbowKit used to be broken and not pass the correct option to wagmi, but they've fixed that now 

Closes RIG-39

### DIFF
--- a/garage/package-lock.json
+++ b/garage/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "garage",
   "version": "0.0.0",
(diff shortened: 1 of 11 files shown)


sanderpick (approved):
Nice, LGTM!

PR #830, opened 2023-05-05
Auto force park Rigs

dtbuchholz:
This PR adds the ability to "auto" force parking a Rig on Blur, OpenSea, and LooksRare. It simply logs the Rig `tokenId`s that are listed while in-flight. Spot checking UIs confirmed the results are accurate per marketplace. The Rigs contract was also updated to use the latest methods from `evm-tableland`.


dtbuchholz:
@sanderpick I ended up just creating a new PR (instead of https://github.com/tablelandnetwork/rigs/pull/790) since I rebased and had some issues with getting the history right. Anyways, this doesn't actually impl / uncomment the force park automation but adds OpenSea and LooksRare APIs to what you started.

sanderpick:
Very cool. So LooksRare API is just open? I'll play around w/ it early next week!

dtbuchholz:
Yeah, no auth for GET reqs but key needed for POSTs. 

I also came across the Rarible aggregator API but didn't play w it at all: https://multichain-api.rarible.org/v0.1/tag/order-controller#operation/getSellOrders. You can look across Rarible, OS, LooksRare, CryptoPunks, ImmutableX, Hen, Objkt, Versum, Teia, X2Y2, Sudo Swap, Fx Hash...no Blur, though.

sanderpick (approved):
@dtbuchholz i just tested this out. works well! I think it's better to leave it in the current state where it just prints the rigs to park vs. doing it auto so that you can verify each one on the marketplaces. takes a bit of time, but way better than starting from zero. I pushed a cleanup commit that just deleted commented code.

dtbuchholz:
cool, sgtm!

PR #835, opened 2023-05-15
Voting Contract

datadanne:
I moved the VotingRegistry contract into the ethereum dir so that we can store table names in the deployment info. Since the VotingRegistry tests use local-tableland I added kind of a hack to ignore them when we run the `npm test` target and added a separate `test-voting` target. We should migrate all tests into using local-tableland probably but that felt a bit out of scope for this PR

Closes RIG-40


dtbuchholz:
Figured I'd share this private on-chain voting implementation by a16z. I'm not saying it should be implemented or anything; it's just a bit related to our idea: https://github.com/a16z/cicada

voting/contracts/VotingRegistry.sol
```diff
@@ -0,0 +1,305 @@
...
+
+    string private constant _FT_SNAPSHOT_PREFIX = "ft_snapshot";
+    uint256 private _ftSnapshotTableId;
+    string private _ftSnapshotTableName;
+
+    string private constant _VOTES_PREFIX = "votes";
+    uint256 private _votesTableId;
+    string private _votesTableName;
+
+    string private constant _ALTERNATIVES_PREFIX = "alternatives";
```
sanderpick, line 36: options?

sanderpick (commented):
Sorry for the crickets here. I finally was able to dig in. Looking good 💪

ethereum/contracts/IVotingRegistry.sol
sanderpick, line 1: Ideally this lives under `interfaces`, but nbd since some files in there are redirects and others aren't. We can clean all that up in a diff PR.

sanderpick (commented):
Looks great! I think we should come up with a plan to test this w/ the whole team. Maybe it's time to setup a live "staging" deployment of the garage on arbitrum testnet / polygon testnet. What do you think? 

We should also point this to the staging branch.

datadanne:
> Looks great! I think we should come up with a plan to test this w/ the whole team. Maybe it's time to setup a live "staging" deployment of the garage on arbitrum testnet / polygon testnet. What do you think?
> 
> We should also point this to the staging branch.

Yup I totally agree, would be great to test this a bit with a larger group. A live staging deployment sounds like a good plan!

ethereum/contracts/IVotingRegistry.sol
datadanne: Ah good catch, my bad!

sanderpick:
cool, i added a linear issue about it

datadanne:
> Looks great! I think we should come up with a plan to test this w/ the whole team. Maybe it's time to setup a live "staging" deployment of the garage on arbitrum testnet / polygon testnet. What do you think?
> 
> We should also point this to the staging branch.

I changed the base to staging + rebased the branch, should be ready to merge if you think it is ready! I can move `IVotingRegistry` to the interfaces dir now too before we merge it if you want me to?

sanderpick:
sweet. I think we should leave it here for now (hold off on merge to staging) until we're in the ship cycle. staging currently has aaron's filecoin work which will be shipping next week.

> I can move IVotingRegistry to the interfaces dir now in this branch if you want me to?

I think it's all good how you have it. We can do some cleanup work on a diff PR at some point (updating all the imports to not be global, and some other nits like that).

sanderpick (approved):

sanderpick:
trying to deploy this to mumbai, but looks like the deploy script is only intended for local-tableland? is that correct?

datadanne:
> the lines are the same, but not like a bunch of code duplication

ah yeah it's because the local `proposal` definition shadows the method named proposal, we could rename the variables or the method if we want to get rid of it

sanderpick:
> > the lines are the same, but not like a bunch of code duplication
> 
> ah yeah it's because the local `proposal` definition shadows the method named proposal, we could rename the variables or the method if we want to get rid of it

i see... makes sense. i think it would be worthwhile to fix... may as well

PR #836, opened 2023-05-19
feat: token reputation

dtbuchholz:
This PR implements token-based reputation. It is inclusive of a draft (unsubmitted) EIP located in `research`, the minimal `ITokenReputation` interface, and associated adjustments to other contracts and tests.

In relation to the EIP, the following provides more detail and can be found at `research/eip-draft_token_reputation.md`:
- This is the core EIP for token-based reputation. It is a minimal interface that simply has staking & unstaking methods along with events; the design is to be as lightweight as possible and hone in on metadata having the "value" and storing reputation using indexed events / data availability.
- There is an ERC-165 interface ID, which _should be_ correct for the supplied token reputation contract—the code to generate this ID was with `type(ITokenReputation).interfaceId`, which is `0x88832242`.
- With a partial goal of EIP acceptance, the majority of the EIP focuses on non-Tableland specific features and includes real-world usage by OpenSea today. [OS disables transfers](https://docs.opensea.io/docs/metadata-standards#disable-trading-for-staked-or-locked-tokens) on a `Stake` event as long as it includes a `tokenId` parameter; it can include additional params and still works as expected (according to the OS support team). The `Unstake` event will re-enable sales.
- At the end of the EIP, two example implementations are provided. The first is generic but at least names the ERC721 token and `baseURI` with Tableland.
- The second example is very Tableland specific (SQLHelpers, TablelandDeployments, gateway query) and tries to sort of justify the Tableland-specific example using the comments to explain what the dynamic metadata does at a high level.
  - This example is fully functional (you can deploy the code) but, perhaps, isn't the "most real" implementation; however, an EIP must be as generic as possible, so this was the ultimate idea with the attempt.
  - The contract creates a pilot sessions like setup and inserts/updates table data; the `tokenURI` queries the Tableland gateway with a `select` statement that forms a basic `attributes` array, calculating `Reputation` with a naive `end_time` minus `start_time` (i.e., no aggregate sum; just a basic example of a row's reputation score).

With the `ITokenReputation.sol` contract, it has been implemented by the `TablelandRigs.sol` contract. This was easier than doing so in `TablelandPilots.sol` as the function signature was more aligned to the core Rigs contract. It was rather straightforward to drop it in, which is a good sign from a usability standpoint (only took a few mins to adjust the exist train / pilot / park methods and tests). Note that this does not change any core contract logic but only method names; external apps like The Garage are affected since the Rigs Garage functions are now different. Lastly, a couple of tests were also added.

Closes [REA-2](https://linear.app/tableland/issue/REA-2/write-preliminary-eip), [REA-3](https://linear.app/tableland/issue/REA-3/implement-generic-contract-interfaces).

### DIFF
--- a/ethereum/contracts/ITablelandRigs.sol
+++ b/ethereum/contracts/ITablelandRigs.sol
@@ -217,18 +217,6 @@ interface ITablelandRigs {
     ) external view returns (ITablelandRigPilots.PilotInfo[] memory);
 
     /**
-     * @dev Trains a single Rig for a period of 30 days, putting it in-flight.
-     *
-     * tokenId - the unique Rig token identifier
-     *
-     * Requirements:
-     *
-     * - `tokenId` must exist
-     * - pilot status must be valid (`UNTRAINED`)
-     */
-    function trainRig(uint256 tokenId) external;
-
-    /**
      * @dev Puts multiple Rigs in training.
      *
      * tokenIds - the unique Rig token identifier
@@ -241,7 +229,7 @@ interface ITablelandRigs {
      * - Values are processed in order
      * - See `trainRig` for additional constraints on a per-token basis
      */
-    function trainRig(uint256[] calldata tokenIds) external;
+    function stake(uint256[] calldata tokenIds) external;
 
     /**
      * @dev Puts a single Rig in flight by setting a custom pilot.
@@ -259,7 +247,7 @@ interface ITablelandRigs {
      * - `pilotId` must be owned by `msg.sender` at `pilotContract` (does not apply to trainer pilots)
      * - Pilot can only be associated with one Rig at a time; parks the other Rig on conflict (does not apply to trainer pilots)
      */
-    function pilotRig(
+    function stake(
         uint256 tokenId,
         address pilotContract,
         uint256 pilotId
@@ -279,26 +267,13 @@ interface ITablelandRigs {
      * - There cannot exist a duplicate value in each of the individual parameters,
      *   except if using a trainer pilot (i.e., trainers aren't unique/owned NFTs).
      * - Values are processed in order (i.e., use same index for each array)
-     * - See `pilotRig` for additional constraints on a per-token basis
+     * - See `stake` for additional constraints on a per-token basis
      */
-    function pilotRig(
+    function stake(
         uint256[] calldata tokenIds,
         address[] calldata pilotContracts,
         uint256[] calldata pilotIds
     ) external;
-
-    /**
-     * @dev Parks a single Rig and ends the current pilot session.
-     *
-     * tokenId - the unique Rig token identifier
-     *
-     * Requirements:
-     *
-     * - `tokenId` must exist
-     * - pilot status must be `TRAINING` or `PILOTED`
-     * - pilot must have completed 30 days of training
-     */
-    function parkRig(uint256 tokenId) external;
 
     /**
      * @dev Parks multiple Rigs and ends the current pilot session.
@@ -312,7 +287,7 @@ interface ITablelandRigs {
      * - Values are processed in order
      * - See `parkRig` for additional constraints on a per-token basis
      */
-    function parkRig(uint256[] calldata tokenIds) external;
+    function unstake(uint256[] calldata tokenIds) external;
 
     /**
      * @dev Allows contract owner to park any Rig that may be intentionally
--- a/ethereum/contracts/ITablelandRigs.sol
+++ b/ethereum/contracts/ITablelandRigs.sol
@@ -1,12 +1,13 @@
 // SPDX-License-Identifier: Unlicense
 pragma solidity >=0.8.10 <0.9.0;
 
-import "./ITablelandRigPilots.sol";
+import {ITablelandRigPilots} from "./ITablelandRigPilots.sol";
+import {ITokenReputation} from "./interfaces/ITokenReputation.sol";
 
 /**
  * @dev Interface of a TablelandRigs compliant contract.
  */
-interface ITablelandRigs {
+interface ITablelandRigs is ITokenReputation {
     // Thrown when minting with minting not open.
     error MintingClosed();
 
@@ -217,7 +218,19 @@ interface ITablelandRigs {
     ) external view returns (ITablelandRigPilots.PilotInfo[] memory);
 
     /**
-     * @dev Puts multiple Rigs in training.
+     * @dev Stakes a single Rig, putting it in-flight with a "stock" pilot.
+     *
+     * tokenId - the unique Rig token identifier
+     *
+     * Requirements:
+     *
+     * - `tokenId` must exist
+     * - pilot status must be valid (`PARKED`)
+     */
+    function stake(uint256 tokenId) external;
+
+    /**
+     * @dev Stakes multiple Rigs, putting them in-flight with a "stock" pilot.
      *
      * tokenIds - the unique Rig token identifier
      *
@@ -227,7 +240,6 @@ interface ITablelandRigs {
      * - `msg.sender` must own the Rig
      * - There cannot exist a duplicate value in `tokenIds`
      * - Values are processed in order
-     * - See `trainRig` for additional constraints on a per-token basis
      */
     function stake(uint256[] calldata tokenIds) external;
 
@@ -242,10 +254,10 @@ interface ITablelandRigs {
      *
      * - `tokenId` must exist
      * - `msg.sender` must own the Rig
-     * - Must be trained & flying with trainer, or already trained & parked
-     * - `pilotContract` must be an ERC-721 contract *or* 0x0 to indicate a trainer pilot; cannot be the Rigs contract
-     * - `pilotId` must be owned by `msg.sender` at `pilotContract` (does not apply to trainer pilots)
-     * - Pilot can only be associated with one Rig at a time; parks the other Rig on conflict (does not apply to trainer pilots)
+     * - Must be currently `PARKED`
+     * - `pilotContract` must be an ERC-721 contract *or* 0x0 to indicate a stock pilot; cannot be the Rigs contract
+     * - `pilotId` must be owned by `msg.sender` at `pilotContract` (does not apply to stock pilots)
+     * - Pilot can only be associated with one Rig at a time; parks the other Rig on conflict (does not apply to stock pilots)
      */
     function stake(
         uint256 tokenId,
@@ -265,7 +277,7 @@ interface ITablelandRigs {
      * - All input parameters must be non-empty
      * - All input parameters must have an equal length
      * - There cannot exist a duplicate value in each of the individual parameters,
-     *   except if using a trainer pilot (i.e., trainers aren't unique/owned NFTs).
+     *   except if using a stock pilot (i.e., stock pilots aren't unique/owned NFTs).
      * - Values are processed in order (i.e., use same index for each array)
      * - See `stake` for additional constraints on a per-token basis
      */
@@ -276,7 +288,19 @@ interface ITablelandRigs {
     ) external;
 
     /**
-     * @dev Parks multiple Rigs and ends the current pilot session.
+     * @dev Unstakes a single Rig and ends the current pilot session.
+     *
+     * tokenId - the unique Rig token identifier
+     *
+     * Requirements:
+     *
+     * - `tokenId` must exist
+     * - pilot status must be `PILOTED`
+     */
+    function unstake(uint256 tokenId) external;
+
+    /**
+     * @dev Unstakes multiple Rigs and ends the current pilot session.
      *
      * tokenIds - the unique Rig token identifiers
      *
@@ -285,14 +309,14 @@ interface ITablelandRigs {
      * - Input array of `tokenIds` must be non-empty
      * - There cannot exist a duplicate value in `tokenIds`
      * - Values are processed in order
-     * - See `parkRig` for additional constraints on a per-token basis
+     * - See `unstake` for additional constraints on a per-token basis
      */
     function unstake(uint256[] calldata tokenIds) external;
 
     /**
-     * @dev Allows contract owner to park any Rig that may be intentionally
+     * @dev Allows contract owner to unstake any Rig that may be intentionally
      * causing buyers to lose gas on sales that can't complete while the
-     * Rig is in-flight.
+     * Rig is in-flight (staking).
      *
      * tokenIds - the unique Rig token identifiers
      *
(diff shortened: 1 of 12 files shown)


awmuncy:
This isn't a comment on the code, so much as a comment on the EIP; it seems weird for a reputation token to be salable. That almost seems to defeat the point of it being a _reputation_ token, no? How should I read that?

sanderpick:
Very good stuff! Finally diving in here.

sanderpick:
> I should probably add these thoughts to the EIP itself, assuming this helps clarifies the EIP's line of thinking! Perhaps the naming convention could be changed to help?

These are great point @dtbuchholz. Def sound worth adding to the EIP

ethereum/contracts/TablelandRigs.sol
```diff
@@ -25,6 +26,7 @@ contract TablelandRigs is
     ERC721AUpgradeable,
     ERC721AQueryableUpgradeable,
     IERC4906,
+    ITokenReputation,
```
sanderpick, line 29: I think this will have to be moved to the end of the list for upgradability

ethereum/contracts/ITablelandRigs.sol
```diff
@@ -216,18 +216,6 @@ interface ITablelandRigs {
...
-     * @dev Trains a single Rig for a period of 30 days, putting it in-flight.
-     *
-     * tokenId - the unique Rig token identifier
-     *
-     * Requirements:
-     *
-     * - `tokenId` must exist
-     * - pilot status must be valid (`UNTRAINED`)
-     */
-    function trainRig(uint256 tokenId) external;
```
sanderpick, line 229: dropping this method, correct?

sanderpick (commented):
Looking great. I left a couple comments.

dtbuchholz:
@sanderpick before I make an "official" EIP PR, I just wanted to double check the draft looks good to go.

ethereum/test/TablelandRigPilots.ts
```diff
@@ -1,5 +1,7 @@
 import { SignerWithAddress } from "@nomiclabs/hardhat-ethers/signers";
 import { loadFixture } from "@nomicfoundation/hardhat-network-helpers";
+import "@nomicfoundation/hardhat-toolbox";
+import "@openzeppelin/hardhat-upgrades";
```
sanderpick, line 4: what do these do here?

research/eip-draft_token_reputation.md
```diff
@@ -37,6 +37,14 @@ These two EIPs also lack the emittance of a `block.number` within their events.
 
 For reference, [OpenSea](https://docs.opensea.io/docs/metadata-standards#disable-trading-for-staked-or-locked-tokens) leverages both of the events noted in EIP-5192 and EIP-5753 to disable trading for staked/locked tokens. There are other accepted events by OpenSea, including the `Stake` and `Unstake` events outlined in this proposal, which _do not_ have an EIP associated with them. Although it is not necessarily the primary motivation of this token reputation EIP, it's been noted to demonstrate how marketplaces are using this EIP's event definitions today. In other words, it is true that the events outlined in this proposal **will disable/enable trading a token on marketplaces**; this proposal _already_ has real-world usage.
 
+### Transferrable vs. Non-Transferrable
```
sanderpick, line 40: Related to the below comment, I think it would be useful to show some examples of selecting reputation w/ SQL for different use cases: 
1. the rigs use case: non-transferrable (select on address)
2. game "points" or similar (build up a character and sell it or something): transferrable (select on token id)

sanderpick (commented):
Looking good. One addition question: Possible to upgrade your mumbai contract and do some manual testing?

@datadanne can you please take a look at the draft EIP as well? 🍻

sanderpick:
@dtbuchholz can you remind me why we are removing the training concept? Is it because it's not compatible with the opensea supported stake events? 

If we drop it, can owners re-pilot with the default pilot? it would be cool if they could because it would be like a badge of early support.

dtbuchholz:
@sanderpick I'm still open to keeping training, but I think it's because we created it to buy us time and don't _really_ need it. From a lore perspective though, I'm split because training is an interesting concept. The most recent changes I made here obviously removed training, but the original PR kept it in there--i.e., there aren't compatibility issues and can still exist with OS Stake events / token rep EIP.

To provide some clarity on the comment about backwards compatibility, this was the original setup for the pilot data (but I should probably add Local Tableland tests to double check it's all good):
- Currently training with trainer pilot == `1`
- Completed training and then re-piloted with a trainer == `2`

If we do drop training, everyone (including the Rigs that never tried to train) gets a stock/default pilot in this current setup and can also re-pilot with it--the UI would have to make changes to account for this, though, along with other fn sig changes. But, I did debate using SQL to determine a "fancy" or "boring" stock/default pilot, which aligns with your early adopters desire. 

So, historical session data stores actively piloted with or re-piloted trainers == `1` or `2`: the early adopters. With training dropped, we _could_ could use SQL to check for sessions with a `1` that completed training (the session never ended) _or_ for any sessions with a `2`--these could get the "fancy" Dispatch stock pilot. We'd probably have to store a `block.number` to check "only sessions prior to block X are valid" here. 

For anything after block X, these Rigs get the "boring" stock/default pilots (the image we use today). That is, the metadata can hold these as attributes and also dynamically render the image based on the SQL response. (This is, in part, why I decided to use `1` as the value for all stock pilots.)

sanderpick:
Makes sense. Thanks for the detail. I'm coming around to your current setup, ie, removed training and make the stock pilot available to everyone. We have a bunch of community rigs that haven't ever been trained... would be a bummer to devalue those by not giving them a cool stock pilot. We could always add some additional metadata to indicate "early adopters" that could be shown w/ a badge or something since all the data is there to make that distinction.

dtbuchholz:
> We have a bunch of community rigs that haven't ever been trained... would be a bummer to devalue those by not giving them a cool stock pilot

@sanderpick a couple of ideas:
1. We put all of these in training, fly for 172800 blocks, and then park them; thus, they'll all have completed training and future-proof for some early adopters setup. We can deploy the changes in this PR to mainnet after this occurs.
2. Create a snapshot of these tokens and store them in a static table on some chain—maybe Polygon or Optimism, which adds to our cross-chain story (i.e., we'd eventually have Rigs tables on Ethereum, Arbitrum One, Filecoin, and Polygon/Optimism). The "early adopters" table. We could choose to do this with _all_ of the early adopter Rigs (not just the community ones) that have completed training, too, to make the query I noted above a bit easier.

Needless to say, this adds a bit of complexity. Related to your other comment—should I deploy the current changes in this PR to the Mumbai contract, or wait until we sift out some of these details before doing so? (I guess that depends on option 1 vs. 2.)

research/eip-draft_token_reputation.md
```diff
@@ -0,0 +1,376 @@
...
+  - Adds `lock` and `unlock` methods, which behave similar to the token reputation proposal's `stake` and `unstake` methods; however, `unlock` lacks the required method caller information, as do the `Lock` and `Unlock` events.
+  - Changes the EIP-5192 `locked` getter to a function named `getLocked` that returns an address instead of boolean; but, getters SHOULD NOT be required for a minimal interface as this assumes contract storage is used to track reputation, which is not always the case (e.g., off-chain metadata / DA based reputation tracking).
+
+These two EIPs also lack the emittance of a `block.number` within their events. This is a key component for enabling reputation in metadata, which is a primary yet OPTIONAL feature of this proposal. For reputation to dynamically change within the metadata, there MUST be block information in the event emittance for off-chain indexing as this is needed to materialize any metadata updates with block-related information. In other words, reputation is most often tracked with a start and end block number that bound the staking/unstaking activities.
```
datadanne, line 36: Isn't the block number implicitly included in all log events? https://github.com/ethereum/go-ethereum/blob/master/core/types/log.go#L31-L56

I think it is, and if that is true it should probably not be included as an event parameter to save gas?

research/eip-draft_token_reputation.md
```diff
@@ -0,0 +1,376 @@
...
+
+For reference, [OpenSea](https://docs.opensea.io/docs/metadata-standards#disable-trading-for-staked-or-locked-tokens) leverages both of the events noted in EIP-5192 and EIP-5753 to disable trading for staked/locked tokens. There are other accepted events by OpenSea, including the `Stake` and `Unstake` events outlined in this proposal, which _do not_ have an EIP associated with them. Although it is not necessarily the primary motivation of this token reputation EIP, it's been noted to demonstrate how marketplaces are using this EIP's event definitions today. In other words, it is true that the events outlined in this proposal **will disable/enable trading a token on marketplaces**; this proposal _already_ has real-world usage.
```
datadanne, line 38: Have we tested that OpenSea does pick up these events automatically? I would guess that they use the event signature to look for relevant events, and if we change the number of event parameters or parameter types the signature changes right? https://docs.soliditylang.org/en/latest/abi-spec.html#events

research/eip-draft_token_reputation.md
```diff
@@ -0,0 +1,376 @@
...
+
+On the contrary, if reputation is tied to the NFT token ID and not a single address, then it actually opens up a new set of opportunities for value creation through **reputation transferability**. Namely, take the snippet above but calculate reputation for token ID `1` instead of address `0x1234...`. In terms of real-world applications, in a gaming, you might earn some reputation that's bound to the token ID, and this unlocks a set of new features that are not available to tokens with a lower reputation. A token-gated workflow. As the token owner, you can choose to sell the token for all of the work you've put in; the tokens with more reputation are more valuable and come with transferrable game state. The owner could even choose to participate in more complex activities, like temporarily allowing another account to own the token (e.g., rent / escrow) and profit from the reputation earned.
+
+### Systematic Example
```
datadanne, line 48: Would it make sense to show the interface definition (`ITokenReputation`) before the example? That feels like a more natural flow to me

research/eip-draft_token_reputation.md
```diff
@@ -0,0 +1,376 @@
...
+
+The approach outlined in this EIP was designed to be as lightweight as possible and only stakes or unstakes a token, emits an event to disable/enable trading, and also allow for recreating token state through the DA layer. It is a generalized implementation and leaves reputation calculation up to the implementor but provides enough context to do so through the defined events.
+
+## Backwards Compatibility
+
+This standard is compatible with [ERC-721](https://eips.ethereum.org/EIPS/eip-721).
+
+## Reference Implementation
+
+Provided is a simple example how token reputation might be implemented with a very minimal `stake()` and `unstake()` implementation that simply emits a `Stake` or `Unstake` event after a token ownership check:
```
datadanne, line 193: This implementation doesn't prevent a token owner from calling `stake` multiple times in a row without calling `unstake`, feels a bit broken and I think I would expect a reference implementation to be functionally complete and work

research/eip-draft_token_reputation.md
```diff
@@ -0,0 +1,376 @@
...
+
+On the contrary, if reputation is tied to the NFT token ID and not a single address, then it actually opens up a new set of opportunities for value creation through **reputation transferability**. Namely, take the snippet above but calculate reputation for token ID `1` instead of address `0x1234...`. In terms of real-world applications, in a gaming, you might earn some reputation that's bound to the token ID, and this unlocks a set of new features that are not available to tokens with a lower reputation. A token-gated workflow. As the token owner, you can choose to sell the token for all of the work you've put in; the tokens with more reputation are more valuable and come with transferrable game state. The owner could even choose to participate in more complex activities, like temporarily allowing another account to own the token (e.g., rent / escrow) and profit from the reputation earned.
```
datadanne, line 46: > In terms of real-world applications, in a gaming, you might earn some reputation

"in a gaming" typo here?

dtbuchholz:
Appreciate all of that feedback! I've adjusted things accordingly and deployed these upgraded contracts to Mumbai.

datadanne:
> Appreciate all of that feedback! I've adjusted things accordingly and deployed these upgraded contracts to Mumbai.

Sweet! I'll take a look at the updates today

sanderpick:
Hey guys, I'm looking through linear and trying to align work that needs shipping. How many weeks out should I put this work?

dtbuchholz:
@sanderpick the contracts/tests should be good to go...and I can make the official EIP repo PR whenever but figured I'd wait to see if the Garage UI ran into hurdles before doing so. Or, I can just do that now if it makes sense to. 

So, should I just merge this to staging and let the voting PR rebase onto this PR's changes? i.e., Linear ship date == the same time as voting?

sanderpick:
That sounds like a good plan. I believe you're still waiting for some feedback on your changes from Daniel (no rush, btw). Makes sense to me to launch these together.

datadanne (commented):
I don't know if I'm lost here but I'm kind of wondering: if we do remove the `owner` parameter from the `Stake` and `Unstake` events (see my comment below), isn't what we are proposing here kind of an 'implementation subset' of what EIP-5753 and EIP-5058 supports? Is it possible that we should be supporting one of those EIPs instead of adding another somewhat similar proposal?

Another reason that makes me think that might be correct is that the actual reputation score isn't part of the interface. We propose that including the reputation score in the NFT metadata is optional, and since the reputation score isn't queryable on-chain there is no 'ethereum-level' interoperability. So it kind of feels more natural to me to have the spec focus on things that are needed for things to be composable on-chain; marketplaces need to know if a token is locked or not to function properly, but they can't use the reputation score if it isn't on-chain or part of the metadata spec

dtbuchholz:
> I don't know if I'm lost here but I'm kind of wondering: if we do remove the `owner` parameter from the `Stake` and `Unstake` events (see my comment below), isn't what we are proposing here kind of an 'implementation subset' of what EIP-5753 and EIP-5058 supports? Is it possible that we should be supporting one of those EIPs instead of adding another somewhat similar proposal?
> 
> Another reason that makes me think that might be correct is that the actual reputation score isn't part of the interface. We propose that including the reputation score in the NFT metadata is optional, and since the reputation score isn't queryable on-chain there is no 'ethereum-level' interoperability. So it kind of feels more natural to me to have the spec focus on things that are needed for things to be composable on-chain; marketplaces need to know if a token is locked or not to function properly, but they can't use the reputation score if it isn't on-chain or part of the metadata spec

fair points. with [eip-5753](https://eips.ethereum.org/EIPS/eip-5753) in mind, i suppose our EIP is basically implementing the same functionality but changing the events and also dropping getter for if a token is locked or not. but, if we were to support 5753 or 5058, then we'd throw out this whole proposal (unless there's a different EIP approach we could take). in full transparency, one goal of our EIP is to simply submit the proposal to help with general Tableland visibility.

datadanne:
> fair points. with eip-5753 in mind, i suppose our EIP is basically implementing the same functionality but changing the events and also dropping getter for if a token is locked or not. but, if we were to support 5753 or 5058, then we'd throw out this whole proposal (unless there's a different EIP approach we could take). in full transparency, one goal of our EIP is to simply submit the proposal to help with general Tableland visibility.

Yeah. I think lockable tokens and having a standard/EIP for that makes sense, and if 5753 and 5058 didn't already exist this EIP would have been great for that. 

I also think it could make sense to have an EIP for a 721-extension for NFT token reputation that focuses solely on that and not the locking/unlocking part (reputation could come maybe be earned in other ways, not just staking NFTs). That feels more composable as well, like implement 721 if you make NFTs, implement 5753 if the tokens are lockable, implement XYZ if the tokens have a reputation score associated with them. But it's not 100% clear to me that NFTs themselves having a reputation tied to them makes sense. For Rigs the holder that stake the Rig earns the reputation and not the Rig, and that feels much more logical to me. It would feel a bit weird to propose a standard that we don't implement haha

dtbuchholz:
closing this as it is no longer being implemented

PR #837, opened 2023-05-29
Garage: Improve piloting UI

datadanne:
A user reported that they weren't able to pilot a rig using a token from the MAX PAIN AND FRENS collection so I looked into it and it turns out that the collection is using token ids larger than `type(uint32).max` (`0xFFFFFF`) which our pilot contract doesn't support because of a packing optimization we made in the pilot info struct (I dunno why they use those IDs, the collection has < 10k tokens but token ids don't start at 0). As I was looking into that I was using the garage as that user and noticed that we weren't filtering out a bunch of garbage tokens from the pilot UI, so I made some tweaks to the piloting UI. 

Changes:
- Filter out airdropped NFTs. The alchemy API does a pretty bad job at filtering out scam NFTs (and doesn't support filtering on ERC721 / ERC1155..), filtering out airdropped NFTs removes most of the scam 1155s
- Filter out ERC721s with tokenId > `0xFFFFFF` because the pilot contract doesn't support them
- Add a fallback image for NFTs without a valid image uri
- Add a fallback way of parsing a token name if the tokenUri is invalid (collection name + tokenId)
- Updated the alchemy sdk so that we can use the newly added `getContractMetadataBatch` sdk method instead of manually calling their api using `fetch`

Before:
image

After:
image

### DIFF
--- a/garage/src/components/FlyParkModals.tsx
+++ b/garage/src/components/FlyParkModals.tsx
@@ -502,6 +502,18 @@ const useFilters = <T,>() => {
   return { filters, toggleFilter, clearFilters };
 };
 
+// The pilot contract packs pilot struct data and requires that the pilot token id
+// is < type(uint32).max
+const MAX_SUPPORTED_PILOT_TOKEN_ID = ethers.BigNumber.from("0xFFFFFFFF");
+
+const tryParseBigNumber = (v: any) => {
+  try {
+    return { valid: true, result: ethers.BigNumber.from(v) };
+  } catch (_) {
+    return { valid: false, result: ethers.BigNumber.from(0) };
+  }
+};
+
 const PickRigPilotStep = ({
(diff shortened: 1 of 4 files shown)


garage/src/components/FlyParkModals.tsx
```diff
@@ -500,6 +502,18 @@ const useFilters = <T,>() => {
   return { filters, toggleFilter, clearFilters };
 };
 
+// The pilot contract packs pilot struct data and requires that the pilot token id
+// is < type(uint32).max
+const MAX_SUPPORTED_PILOT_TOKEN_ID = ethers.BigNumber.from("0xFFFFFFFF");
```
sanderpick, line 507: 👍

sanderpick (approved):
LGTM!

PR #864, opened 2023-06-27
feat: migrate to upgraded `wagmi`, `rainbowkit`

dtbuchholz:
With `wagmi` moving from `ethers` to `viem`, the `useSigner` hook no longer works. Thus, a custom hook has been created, along with other minor changes due to the dep bump/migration. Also, WalletConnect now requires an API key—it's a `projectId` variable used in the `wagmi` setup and is now accessible as `VITE_WALLET_CONNECT_PROJECT_ID` (env var); this has been added within Vercel as well.

### DIFF
--- a/garage/src/hooks/useTablelandConnection.ts
+++ b/garage/src/hooks/useTablelandConnection.ts
@@ -1,7 +1,7 @@
-import { Database, Validator } from "@tableland/sdk";
+import { Database, Validator, helpers } from "@tableland/sdk";
 import { chain } from "../env";
 
-const db = Database.readOnly(chain.id);
+const db = new Database({ baseUrl: helpers.getBaseUrl(chain.id) });
 const validator = new Validator(db.config);
 const data = { db, validator };
 
--- a/garage/src/pages/RigDetails/index.tsx
+++ b/garage/src/pages/RigDetails/index.tsx
@@ -162,6 +162,7 @@ export const RigDetails = () => {
   const { validator } = useTablelandConnection();
 
   const { data: contractData, refetch } = useContractReads({
+    allowFailure: false,
     contracts: [
       {
         address: as0xString(contractAddress),
@@ -285,8 +286,8 @@ export const RigDetails = () => {
                   <RigHeader
                     {...MODULE_PROPS}
                     rig={rig}
-                    owner={(owner as any)?.result}
-                    tokenURI={(tokenURI as any)?.result}
+                    owner={owner}
+                    tokenURI={tokenURI}
                     userOwnsRig={userOwnsRig}
                     currentBlockNumber={Number(currentBlockNumber)}
                     refresh={refresh}
@@ -313,8 +314,8 @@ export const RigDetails = () => {
                   <RigHeader
                     {...MODULE_PROPS}
                     rig={rig}
-                    owner={(owner as any)?.result}
-                    tokenURI={(tokenURI as any)?.result}
+                    owner={owner}
+                    tokenURI={tokenURI}
                     userOwnsRig={userOwnsRig}
                     currentBlockNumber={Number(currentBlockNumber)}
                     refresh={refresh}
@@ -327,7 +328,7 @@ export const RigDetails = () => {
                   onOpenParkModal={onOpenParkModal}
                   onOpenPilotModal={onOpenPilotModal}
                   onOpenTrainModal={onOpenTrainModal}
-                  chainPilotStatus={pilotInfo?.status === "success" ? 1 : 0}
+                  chainPilotStatus={pilotInfo?.status}
                   {...MODULE_PROPS}
                 />
                 <FlightLog rig={rig} nfts={nfts} {...MODULE_PROPS} />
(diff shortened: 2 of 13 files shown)


dtbuchholz:
@datadanne not sure what the next step is to resolve the issues; it doesn't read/render data correctly and throws when requesting Rig info. That is, the wallet connection should be working properly, but the `wagmi` hooks used when reading/writing to contracts might be the root cause. The `wagmi` upgrade drops `BigNumber`s to standard `bigint`s, which are used in hooks such as in `useContractWrite` and `useContractRead`.

I took a stab in the most recent commit to migrate `ethers` types to those needed by `wagmi` in hopes it would help with resolve things. However, feel free to drop that commit and/or take over this PR...it ended up being a bit more involved more than I anticipated, past the rather standard `wagmi` migration!

garage/src/pages/RigDetails/index.tsx
```diff
@@ -199,7 +199,8 @@ export const RigDetails = () => {
   const userOwnsRig = useMemo(() => {
     return (
       !!actingAsAddress &&
-      actingAsAddress.toLowerCase() === owner?.toLowerCase()
+      actingAsAddress.toLowerCase() ===
+        ((owner as any)?.toLowerCase() as string)
```
dtbuchholz, line 203: This is one of the obvious issues—I was running into issues about `toLowercase` being used incorrectly. E.g., see the logs upon requesting Rig info: https://rigs-garage-dj48r6f04-tableland.vercel.app/rigs/356

garage/src/pages/RigDetails/index.tsx
```diff
@@ -287,10 +285,10 @@ export const RigDetails = () => {
                   <RigHeader
                     {...MODULE_PROPS}
                     rig={rig}
-                    owner={owner}
-                    tokenURI={tokenURI}
+                    owner={(owner as any)?.result}
```
dtbuchholz, line 288: And then this was in attempt to resolve `Type '{ error: Error; result?: undefined; status: "failure"; }' is not assignable to type 'string'` but looks wrong

garage/src/pages/RigDetails/index.tsx
```diff
@@ -329,7 +327,7 @@ export const RigDetails = () => {
                   onOpenParkModal={onOpenParkModal}
                   onOpenPilotModal={onOpenPilotModal}
                   onOpenTrainModal={onOpenTrainModal}
-                  chainPilotStatus={pilotInfo?.status}
+                  chainPilotStatus={pilotInfo?.status === "success" ? 1 : 0}
```
dtbuchholz, line 330: i'm realizing i made some incorrect assumptions, so this is wrong as well. the `wagmi` interface change now returns a `success` or `failure` value for a `status` key. with my changes, `pilotInfo?.status` was reading this value instead of something from the underlying `pilotInfo?.result`.

so this should maybe be `Number(pilotInfo?.result)`...but i'm still not sure how to get the correct value from the `wagmi` hook responses

datadanne:
> @datadanne not sure what the next step is to resolve the issues; it doesn't read/render data correctly and throws when requesting Rig info. That is, the wallet connection should be working properly, but the wagmi hooks used when reading/writing to contracts might be the root cause. The wagmi upgrade drops BigNumbers to standard bigints, which are used in hooks such as in useContractWrite and useContractRead.
> 
> I took a stab in the most recent commit to migrate ethers types to those needed by wagmi in hopes it would help with resolve things. However, feel free to drop that commit and/or take over this PR...it ended up being a bit more involved more than I anticipated, past the rather standard wagmi migration!

Thanks for doing all of this! Looks great, I think the only thing we need to fix is the useage `useContractReads`. They've changed the return type from just data to something like `{ error?: Error; result?: T; status: string; }` unless the `allowFailure` parameter is set to false, so we can just set that and we don't have to unwrap the `result` everywhere. For some reason type inference for the return data isn't working after the wagmi upgrade though with they way we've defined the contract abis, I'll look into that today so that we can get this PR ready!

dtbuchholz:
ah nice, that made it easy! (i can't review since it's my PR but lgtm)

datadanne (approved):
Nice!

datadanne:
> ah nice, that made it easy! (i can't review since it's my PR but lgtm)

@dtbuchholz approved! we should probably deploy this asap right since walletconnect will stop working today if we don't?

(side note but kind of random that I can push commits to a PR that you opened, approve it as a reviewer and then merge it if I want to haha. makes the "require at least one approval" check pretty pointless ^^)

garage/src/hooks/useTablelandConnection.ts
```diff
@@ -1,7 +1,7 @@
-import { Database, Validator } from "@tableland/sdk";
+import { Database, Validator, helpers } from "@tableland/sdk";
 import { chain } from "../env";
 
-const db = Database.readOnly(chain.id);
+const db = new Database({ baseUrl: helpers.getBaseUrl(chain.id) });
```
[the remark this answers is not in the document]
dtbuchholz: iirc `readonly` was a bit confusing because you'd specify a single chain...but reads can be cross chain. so, that was dropped and just keeps it generalized

dtbuchholz:
Cool, will merge. And lol yeah it's a bit odd...I suppose a rule like "no one made a commit on this branch/PR" should be used, if that's a thing

PR #869, opened 2023-06-30
Update staging branch with main

datadanne:
staging is out of sync with main because we merged https://github.com/tablelandnetwork/rigs/pull/864 directly to main


datadanne:
I wonder if the update branch-button works for this PR? it really shouldn't since that would merge all of staging into main which would automatically deploy the code, hopefully the branch protection rules wouldn't allow it if someone clicked it

asutula (approved):

PR #878, opened 2023-07-04
Garage: add support for staging deployment

datadanne:
Updates the code to support multiple deployments. Staging will use the `polygon-mumbai` deployment from the rigs package, and production uses the mainnet deployment.

I've added `VITE_APP_ENV`  env var defs to vercel. All preview deployments (PRs & staging branch) will be using the staging deployment, and prod will use production

Closes RIG-49


sanderpick (approved):
beauty

PR #879, opened 2023-07-04
Garage: add multi chain support

datadanne:
Adds support to the garage for being connected to multiple chains. Needed when using the admin page to write to tables that are not on mainnet, and will be needed for voting

NOTE: The base of this PR is https://github.com/tablelandnetwork/rigs/pull/878 and not staging, because it depends on the work in 878. Will change base as soon as that one is merged

Closes RIG-51


sanderpick (dismissed):
nice! lgtm

datadanne:
> nice! lgtm

sweet! will need a re-approval since the base changed to staging

sanderpick (approved):

PR #882, opened 2023-07-05
Ship 8b

sanderpick:
https://linear.app/tableland/project/rigs-filecoin-integration-10012ecc0ad8


asutula (dismissed):

sanderpick:
I'm seeing an error in the garage deployment:
```
Uncaught (in promise) Error: ALL_ERROR: running read statement: parsing result to json: executing query: no such column: renders_cid
SELECT
    *,
    'ipfs://'||renders_cid||'/'||rig_id||'/'||image_thumb_name as "thumb",
    'ipfs://'||renders_cid||'/'||rig_id||'/'||image_full_name as "image"
  FROM (
    SELECT
      rig_id,
      cast(rig_id as text) as "rigId",
      json_object(
        'contract', pilot_contract,
        'tokenId', pilot_id
      ) as "pilot",
      'piloted' as "action",
      start_time as "timestamp"
    FROM pilot_sessions_80001_4078
    WHERE end_time IS NULL 
    UNION
    SELECT
      rig_id,
      cast(rig_id as text) as "rigId",
      null as "pilot",
      'parked' as "action",
      end_time as "timestamp"
    FROM pilot_sessions_80001_4078
    WHERE end_time IS NOT NULL 
  )
  JOIN lookups_314159_10
  ORDER BY timestamp DESC
  LIMIT 40
  OFFSET 0
    at bf (index-e0345121.js:541:960)
    at Wk.all (index-e0345121.js:542:1710)
```

@asutula is taking a stab at it but @datadanne we might need your assistance

datadanne:
> I'm seeing an error in the garage deployment:
> 
> ```
> 
> Uncaught (in promise) Error: ALL_ERROR: running read statement: parsing result to json: executing query: no such column: renders_cid
> 
> SELECT
> 
>     *,
> 
>     'ipfs://'||renders_cid||'/'||rig_id||'/'||image_thumb_name as "thumb",
> 
>     'ipfs://'||renders_cid||'/'||rig_id||'/'||image_full_name as "image"
> 
>   FROM (
> 
>     SELECT
> 
>       rig_id,
> 
>       cast(rig_id as text) as "rigId",
> 
>       json_object(
> 
>         'contract', pilot_contract,
> 
>         'tokenId', pilot_id
> 
>       ) as "pilot",
> 
>       'piloted' as "action",
> 
>       start_time as "timestamp"
> 
>     FROM pilot_sessions_80001_4078
> 
>     WHERE end_time IS NULL 
> 
>     UNION
> 
>     SELECT
> 
>       rig_id,
> 
>       cast(rig_id as text) as "rigId",
> 
>       null as "pilot",
> 
>       'parked' as "action",
> 
>       end_time as "timestamp"
> 
>     FROM pilot_sessions_80001_4078
> 
>     WHERE end_time IS NOT NULL 
> 
>   )
> 
>   JOIN lookups_314159_10
> 
>   ORDER BY timestamp DESC
> 
>   LIMIT 40
> 
>   OFFSET 0
> 
>     at bf (index-e0345121.js:541:960)
> 
>     at Wk.all (index-e0345121.js:542:1710)
> 
> ```
> 
> 
> 
> @asutula is taking a stab at it but @datadanne we might need your assistance


Ah, did we publish a new lookups table that doesn't have that column? 

If that's what's causing it the queries will have to be rewritten, but as a quick fix to unblock the deployment you could just hardcode it to use the old lookups table in env.ts instead of reading from the rigs package

asutula:
Yea @datadanne that is the problem. I created an issue about it here https://linear.app/tableland/issue/RIG-52/update-garage-to-use-new-rig-image-cids-and-lookups-table. Good call about hardcoding the old table as a work around. Since it's late in the day here, let's just see what you think about my branch I handed off to you in that issue, and if it seems tricky still by my tomorrow morning, we can do the hardcode workaround.

asutula:
Ok this is looking good.

asutula (approved):

sanderpick:
@datadanne i noticed a couple things on the live deployment: missing owner address on rig pages and a little text overflow issue on the deal popup

Screenshot from 2023-07-06 08-30-55
Screenshot from 2023-07-06 08-29-25

sanderpick:
oh, I forgot that the previews are pointing to the mumbai deployment. is that correct?

datadanne:
> oh, I forgot that the previews are pointing to the mumbai deployment. is that correct? 

Yeah. The missing owner info is because we deployed a new contract to mumbai recently so no tokens are minted but the tables are populated with data

sanderpick:
Is the "raw metadata" button supposed to work on the staging deploys?

datadanne:
> Is the "raw metadata" button supposed to work on the staging deploys?

No right now it won't, it gets the URL by calling the `tokenURI` method on the contract and we don't get that data right now because we use a multicall to call multiple read functions in one RPC request. `ownerOf` is one of read calls and that one reverts for unminted tokens

asutula (approved):

PR #900, opened 2023-07-18
Run VotingRegistry tests in CI

datadanne:


viewer/package.json
```diff
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "nuxt",
     "start": "nuxt start",
-    "generate": "nuxt generate",
+    "generate": "NODE_OPTIONS=--openssl-legacy-provider nuxt generate",
```
datadanne, line 8: This isn't an ideal solution but I think basically all dependencies (nuxt, vue etc) would have to be updated to avoid this. I tried updating them with `npm audit fix --force` but that broke the build. I tried fixing everything that broke and got to a point where the build was successful but the generated app didn't render any rig images at all so I gave up on that. @asutula if you think it would be easy for you to upgrade the deps to avoid this that might be worth it?

asutula: I think this is fine for now. Can't really afford the context switch right now, it would be severe since I haven't looked at this stuff in months. If this workaround allows things to work, let's go with it and log an issue to update the deps at some point.

asutula: To clarify, the `viewer` web app is just run locally as a development tool, so no big risks involved.

asutula (approved):

PR #903, opened 2023-07-19
VotingRegistry: Add voting systems

datadanne:
To future proof the interface and tables so that we don't have to migrate any data if we want to support more voting systems in the future


datadanne:
@sanderpick could you create a new staging deployment of the contract & all tables once this PR is approved and merged (feel free to merge if you approve it so that you don't have to wait for me merging it)? I wanna do some final testing in the staging env + create a proposal that more people can try voting on to see that everything works correctly. Related to that, can you grant my dev wallet insert permissions on the ft_rewards table for the staging deployment? That way I can setup a bunch of test wallets that anyone that wants to try voting can use that has some mumbai matic and FT, much easier to give ppl FT using the rewards table vs minting rigs & piloting

sanderpick (approved):

PR #911, opened 2023-08-08
Garage: voting fixes

datadanne:
Changes:
- Adds explanation to why the user isn't eligible to vote (you had 0 ft at the time of the snapshot)
- Updates the proposal page to refresh the data & display a succes toast after the user has voted
- Adds a tooltip to the Voting reward text that explains that rewards are distributed after the proposal ends
- Updates the vote button-title to 'Update Vote' if the user has already voted

+ a few small bug fixes


sanderpick (dismissed):
LGTM!

sanderpick (approved):

PR #912, opened 2023-08-09
Ship voting arbitrum

sanderpick:
Deploys all voting tables and the voting registry to Arbitrum. The FT rewards table was re-deployed to Arbitrum as well.


ethereum/hardhat.config.ts
```diff
@@ -133,7 +158,7 @@ const config: HardhatUserConfig = {
 
 interface RigsTablesConfig {
   tablelandPrivateKey: string | undefined;
-  tablelandAlchemyKey?: string;
```
datadanne, line 136: I think this change breaks the `deployRigs` and `deployRigsAndPilots` scripts since they still refer to the alchemy key

datadanne (dismissed):

ethereum/hardhat.config.ts
sanderpick: pushed a fix

PR #918, opened 2023-08-25
Garage: update ft checker vercel function to support both a single wallet and multiple wallets

datadanne:
Closes RIG-65


sanderpick (approved):
lgtm!

PR #922, opened 2023-09-19
Garage: update admin page to check admin permissions

datadanne:
Adds admin permission checks for proposals & missions. I don't think there is an api for checking if an account has been granted permission to write to a table using `GRANT`, so we can't check if the account can give out ft rewards.

image

Closes RIG-68


sanderpick (dismissed):
LGTM!

sanderpick (approved):

PR #930, opened 2023-09-26
Unbreak rigs cli

datadanne:
by removing a leftover TODO comment from a sql statement

### DIFF
--- a/pkg/storage/local/impl/store.go
+++ b/pkg/storage/local/impl/store.go
@@ -111,11 +111,11 @@ const (
 		drop table if exists rigs;
 		drop table if exists rig_parts;
 		drop table if exists rig_deals;
-		// TODO: Investigate what we want here.
-		// drop table if exists cids;
-		// drop table if exists table_names;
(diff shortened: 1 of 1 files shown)


pkg/storage/local/impl/store.go
```diff
@@ -111,10 +111,6 @@ const (
 		drop table if exists rigs;
 		drop table if exists rig_parts;
 		drop table if exists rig_deals;
-		// TODO: Investigate what we want here.
-		// drop table if exists cids;
-		// drop table if exists table_names;
-		// drop table if exists txns;
 	`
```
asutula, line 114: Can we just move the comments outside of the string literal so we don't loose that TODO?
```suggestion
	`
	// TODO: Investigate what we want here.
	// drop table if exists cids;
	// drop table if exists table_names;
	// drop table if exists txns;
```

asutula (changes requested):

asutula (approved):
Great, thanks!

PR #931, opened 2023-10-18
Garage: Add FT leaderboard

datadanne:
image


sanderpick (approved):
Nice! Looks good to me.

PR #969, opened 2023-11-07
Garage: Fix listing of owned NFTs when piloting a Rig

datadanne:
Removes the alchemy AIRDROPS filter from useOwnedNFTs, something muyt have changed in alchemy's api because this filter used to work well but now it filters out a lot of non-spam NFTs


sanderpick (approved):

sanderpick:
nice! thanks for jumping in @datadanne

PR #1016, opened 2024-02-17
Garage: Support piloting trained rigs with the default trainer

datadanne:
Updates the UI pilot rig flow to support clicking the pilot rigs button without selecting a custom pilot, and it will default to the trainer pilot if no pilot is set. Visual changes:

Adds a small note to the pilot modal that you can pilot with the default trainer pilot or a custom pilot:
image

Shows an error if the user tries to pilot a rig that is already in-flight with the default trainer:
image
This would essentialy be a no-op so hopefully no one tries to do this, but since the contract doesn't support that we need to handle the error instead of failing silently


dtbuchholz (approved):

sanderpick (approved):
lgtm!

PR #1020, opened 2024-12-21
feat: use simplehash api for blur rigs

dtbuchholz:
Switches to Simplehash's API for active listing, filterable by marketplace: https://docs.simplehash.com/reference/listings-by-contract-1


sanderpick (approved):
:rocket:

